PR #318, opened 2026-06-30
City grid orientation lock (PAR-246)

### LINKED ISSUE #272: City grid orientation lock
It is common in some cities to orient the map with the city's grid matching the cardinal directions. We should define a hardcoded list of cities with rigid street grids and add a bearing/rotation snap that locks the map rotation to the grid.  <br>  <br>NYC is a good city to test with, city is on a 29° angle

Philly is also gridded on an angle but I'm unsure of the angle.

If we do this, we should snap the angle within a few degrees of closeness, matching Mapbox's default behavior when snapping to north up. 

This feature should be configurable in map settings, default on.

alexwohlbruck:
Implements [PAR-246](https://linear.app/parchment-maps/issue/PAR-246/city-grid-orientation-lock) — snap the map's rotation to a city's street grid (and to north), the way many cities are mentally oriented to their grid rather than true north.

## What it does

When you finish rotating the map near a grid-aligned bearing, it clicks into alignment — the same feel as Mapbox's default north-up snap, but aware of angled city grids (e.g. Manhattan's 29°). Two independent, configurable behaviors in **Settings → Behavior → Map rotation**:

- **Snap to north** *(default on)* — snap upright when a rotation ends near north.
- **Snap to street grid** *(default North up only)* — `Off` · `North up only` · `All 90° rotations`. Active only when the map is centered over a known gridded city.

Both are evaluated together on each rotate-end and snap to whichever aligned bearing is closest (within 7°, matching Mapbox's `bearingSnap`).

## Implementation notes

- **Verified dataset** — `gridded-cities.ts` holds 44 cities whose bearings were each measured from OpenStreetMap street geometry (and cross-checked against documented sources / "henge" data). Cities that measured as cardinal or have no coherent grid were dropped, since the north-up snap already covers cardinal grids.
- **Native `bearingSnap` disabled** — the engines cache `bearingSnap` inside the gesture handler at construction, so it can't be toggled at runtime. We set `bearingSnap: 0` and reimplement north + grid snapping in `map.service` (`snapRotation`), reading the settings live so both toggles take effect without a reload.
- **Gesture-only** — strategies emit `rotateend` only for user gestures (guarded on `originalEvent`), so programmatic camera moves and our own `easeTo` don't loop.
- Pure helpers (`grid-orientation.ts`) are unit-tested (18 tests); 4-fold grid symmetry is handled by folding mod 90.

## Testing

- `vue-tsc` clean, full suite green (690 tests, incl. 18 new for the snap helpers).
- Manually verified in the dev preview against several cities.

Closes #272


alexwohlbruck:
Closing: the PAR-246 work is already integrated into `dev` (feature via #316, plus the settings-move commit pushed directly to dev). This PR was opened against `main` by mistake and is mis-scoped (pulls in unrelated dev history). No separate PR needed.

PR #410, opened 2026-08-22
The portolan transit map in parchment: tiles, service time, and bullets that match

alexwohlbruck:
Parchment renders portolan's transit map with portolan's own styling, from
portolan's tiles. Thirty-six commits; most of the later ones exist because
the first version was watched on a real map and was wrong in ways only a
screen shows.

## The network

- **Standard chunked MVT**, one vector source per pyramid, **mounted by
  viewport** — loading all 89 up front froze the UI for seconds on first
  paint.
- Runs on **both engines**. The MapLibre fork
  (`github:alexwohlbruck/maplibre-gl-js`) gives variable `line-offset` and
  `symbol-anchor-offset`, so junction ramps ease between slots and
  caterpillars hang off an anchor vector. Mapbox degrades those two rather
  than taking the transit map down with it.
- Junction transitions hydrate off **our own sources**, not on the whole
  style settling: `isStyleLoaded()` is never true with 113 sources, and the
  sweeps were starving.
- One failing overlay no longer takes the whole map with it.

## Service time

A control under the Transit group, live and ticking or any weekday hour.
The network filters to that instant — and **route detail isolates by
narrowing portolan rather than drawing over it**, so an opened route shows
the path as it *is*: the stretches it does not run at this hour are gone,
and their stops and labels with them, because the stations answer the same
per-route question the segments do.

## Bullets and labels, matched to the map

The panel and the map draw the same station, so they must agree.

- **Order** is portolan's `sortBullets`, ported line for line, with
  fixtures read out of real builds — Columbus Circle is `A C · B D · 1 2`,
  which is what the MTA prints.
- **Style** is portolan's: the eight curated outlines, the pad each needs,
  and glyph colour by luminance instead of a flat white that printed white
  on the MTA's yellow. The curated shape, colour and label now reach the
  panel through `routes.json`.
- **Station names** letter themselves the way the basemap letters its own —
  white on dark, black on light, following a theme change. This took five
  attempts; the tests now drive the real style builders in every basemap
  and theme, with a live overlay injected, because reading the style was
  the thing that kept being subtly wrong. On Mapbox the basemap is an
  import and exposes nothing, so its `lightPreset` is asked directly.
- **Wrap counts** come from the renderer's own glyph advances. A canvas
  measures a substituted face, and a few per cent decides whether the
  bullet strip lands on the name or a line clear of it.

## Stop pages

A clicked station opens **the same place as the same platform clicked on
the basemap** — `/place/node/1683730419`, not a second page about one stop
keyed by a different vocabulary. Header bullets are links, and dim when the
board has nothing coming, with a tooltip that says so.

---

Depends on portolan `rebuild-onestop` (alexwohlbruck/portolan#16) and
barrelman `portolan-integration` (alexwohlbruck/barrelman#54); merge in
that order.

**Not verified on screen.** The automation browser on the dev box stopped
compositing partway through this work — no `requestAnimationFrame`, no
canvas — so the visual behaviour here is argued from tests and from data
read back out of the tiles, not from looking at it. The label-colour rule
in particular was shipped wrong four times for exactly that reason, which
is why it is now tested against the real style builders rather than
asserted.


alexwohlbruck:

PR #422, opened 2026-08-31
MapLibre POI markers, Barrelman 3D buildings, and plan-view depth

alexwohlbruck:
Seventeen commits of MapLibre map work, in three groups.

## POI markers and icons

* POI badges are built at the size they render, not twice it — the engine was point-sampling them back down and throwing half the edge away, so a disc came out with a stair-stepped rim beside a search-result pin the browser had drawn cleanly. The soft glow no longer ends in a faint square.
* The badge ring gets its own colour, a shade darker than the icon at night, instead of the near-white the glyph wears.
* Search-result pins are drawn with the same plate, glyph and ring as the places already on the map, so a café you searched for and the café under it are visibly the same thing.
* Search-result labels and the saved-place outline follow the theme rather than being stuck in daylight colours.
* Roughly half the icons in the OSM preset schema name an icon set we don't ship, and every one of them silently became a pin. A place now inherits from what it is a kind of, so a taqueria gets the restaurant glyph.
* Minor roads are drawn wider — the map gave up colour for road hierarchy but kept weights tuned for a map that still had it.

## 3D buildings from Barrelman

* Buildings come from Barrelman's own source, with roof colours where OSM records them, and tiles are fetched from `/tiles` at the host the integration stores.
* Buildings with modelled parts are drawn once instead of twice. OSM maps a detailed building as an outline plus separate pieces holding the real heights, and both were being drawn — the pieces at true height, the outline at a default one, flickering against each other. The outline is now dropped server-side via `hide_3d`.
* The basemap's own buildings stay until Barrelman is actually serving; the roofline stand-in points at the same buildings it outlines; the ground buries the basement once the camera tilts.
* Building paint carries hue at any lightness across the whole spectrum. The tint had been scaled by departure from the tile colour's own luminance, which a dark colour has little of however saturated — so maroon and bottle green barely tinted while pale colours tinted strongly, and because it was measured against luminance the palette collapsed onto a blue-yellow axis (dark red facades came out blue at night).
* Dark-mode sand is sand-coloured rather than cyan.

## Plan-view depth, and the tint level

The last three commits:

* **Tint level.** `BUILDING_TINT` drops ~35% (light 34 → 22, dark 46 → 30). The tint is there to tell one building from its neighbour; at the old strength a street of brick read as a wash of colour over the land rather than as buildings standing on it.
* **Clip planes.** The plan view fakes an orthographic camera by narrowing the FOV to 0.5°, which throws the camera ~115 screen heights out. MapLibre pins the near plane at `height / 50` whatever the FOV, so the far-to-near ratio hit ~5800:1 and the depth buffer spent its precision on empty air in front of the camera — a quantisation step tens of pixels deep at the map plane. Trees shattered into light and dark wedges and roofs fought the walls under them. `updateClipPlanes` now closes the planes to `0.5×`/`2.0×` the camera throw when flat on and hands them back to MapLibre at any pitch, taking the ratio to 4:1.

  Widening the FOV instead was the obvious alternative and is the wrong lever: precision scales with the square of the camera distance, so the ~5° needed would put visible walls on every building at the edge of the screen — the flatness the plan view exists for.

  Note this reaches `map._camera.transform`, falling back to `map.transform`. The fork restructured `Map` so it no longer extends `Camera`; reaching only for `map.transform` gets `undefined` and the override silently never runs.
* **`window.__map`** in dev builds, which is what made the above diagnosable.

## Verification

Typecheck clean, 1661 web tests pass. The clip-plane fix was checked against the running preview in headless Chromium: flat-on reads `nearZ` 40,107 / `farZ` 160,427 (ratio 4), and it survives pan, zoom and real drag gestures — the handler path clones and re-applies transforms, so that needed checking specifically. Tilting to 45° restores `autoCalculateNearFarZ`, returning to flat re-applies the override. Screenshots at z19.4 flat-on over First Ward Park show solid crowns.

MapLibre engine only — Mapbox Standard doesn't use the ortho trick.


alexwohlbruck:

PR #436, opened 2026-09-02
Globe view on MapLibre

alexwohlbruck:
Adds globe support to the MapLibre engine and wires it into the existing map
projection setting, which was Mapbox-only until now.

MapLibre projects onto a flat Mercator plane or onto a sphere, and nothing
else, so its picker offers those two while Mapbox keeps the full list.
`ENGINE_PROJECTIONS` records which engine can draw what; the command store
filters its options through it, which covers both the command palette and the
settings page since they read the same source. Switching engines while on a
Mapbox-only projection falls back to Mercator, so the setting never names a
shape the map is not drawing.

### The globe is not a sphere at every zoom

Both engines ease the globe into Mercator as you zoom in — MapLibre across
zoom 11–12, Mapbox across 5–6 — so a globe map is a flat map everywhere a
street is legible. Two things had to follow that rather than the setting:

* **The orthographic plan view.** Narrowing the field of view to 0.5° retreats
  the camera ~115 screen heights, which on a sphere leaves the sphere a speck.
  It now stands down while the globe is on screen and comes back with the flat
  map. Reading MapLibre's own `isGlobeRendering` off the transform does not
  work here: that flag is written during render, so an event handler reads the
  previous frame, and a zoom that ends past the boundary is never reconsidered.
  It is derived from zoom instead.
* **Panel padding.** The map leaves room so a searched place does not land
  under an open panel, which moves the focal point off the middle of the
  viewport. A flat map covers the canvas edge to edge and hides that; a globe
  is a discrete object with an obvious centre, and a 400px panel left it
  sitting 200px right of the middle. Padding is dropped while a sphere is on
  screen and restored when it flattens — nothing is lost, since both engines
  are flat long before the zoom where a panel matters. This fixes the shipped
  Mapbox globe too, not just the new one.

### Atmosphere

MapLibre scatters the globe's halo from the style's `light`, which the strategy
keeps pointed at the real sun so buildings cast their shadows the right way. On
a globe that lights one limb and darkens the other — a second, cruder
terminator under the day/night layer's own tuned twilight ramp. `atmosphere-blend`
is off. Building shadows are untouched.

With no halo, MapLibre draws nothing outside the sphere, so there is a backdrop
behind it: the basemap's own land tone by day, its night sky by dark.

### Verification

Typecheck clean, 1698 tests pass (4 new, in `map-style.test.ts`). Measured on
the running app rather than inferred:

| state | projection | plan-view FOV | padding | centre vs. midpoint |
|---|---|---|---|---|
| zoom 2.5, globe | `{type: 'globe'}`, sphere rendering | 36.87° (default) | zeroed | 614 vs. 613.5 |
| zoom 16, globe selected | flattened to Mercator | 0.5° + clip planes | restored | — |

The evaluated sky property reads `atmosphere-blend: 0`, not just the stylesheet.

The table above was measured before the rebase onto #431. Since then the dev
machine's barrelman tile source has been returning 502, so the map never
reaches `load`, `isMapReady` stays false, and `updateMapPadding` early-returns
— which makes the padding path (the app's own visible-area watcher included)
impossible to exercise there. The post-rebase checks are therefore static:
typecheck, the full suite, and that the merge kept `toContainerRect` at all
three call sites. The padding behaviour itself wants one more look on a machine
with tiles. No screenshot for the same reason.

### Two things worth knowing

* **Padding snaps rather than fades** at the crossing, so zooming in through
  zoom 12 with a wide panel open shifts the map by half the panel width in one
  frame. Scaling padding by the transition instead of switching on it would
  smooth this out; both engines expose the ramp. Left out as it is a contained
  follow-up.
* **Rebased on top of #431**, which fixed a separate cause of the same
  off-centre symptom — obstruction bounds are viewport-space while the map
  container is a flex sibling of the sidebar, so the sidebar was counted twice.
  The two compose: that fix corrects the padding value, this one drops padding
  altogether while a sphere is on screen. `effectiveMapPadding` wraps
  `calculateMapPadding`, so it inherits `toContainerRect` at every call site.
* The projection setting already defaults to `globe`, so existing MapLibre
  users will see the globe when zoomed out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 * The map keeps up with the sidebar as it is dragged or collapsed, resizing frame by frame rather than snapping once the panel has settled
 * Centring on a place, or fitting a route, no longer aims too far to the right when a panel is open over the map. The gutter the map was told to leave on the left included the sidebar's width a second time, even though the map never extended under it — worse the wider the sidebar
+* The globe sits in the middle of the map when a panel is open. That same gutter moves the point the map is centred on off the middle of the screen, which a flat map hides and a globe cannot, so it is dropped while the globe is on screen
 
 ## [0.8.3] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+* The MapLibre engine can now draw the map on a globe. The map projection setting applies to both engines, offering flat and globe on MapLibre and the full list on Mapbox
+
 ### Changed
 
 * The desktop sidebar has been rebuilt. Rows sit on an even grid at full contrast, the current one is marked by a single raised chip that slides between them, and collapsing or expanding the rail now glides instead of snapping — with the map resizing in step
--- a/web/src/components/map/Map.vue
+++ b/web/src/components/map/Map.vue
@@ -170,6 +170,24 @@ watch(
   outline: none;
 }
 
+/*
+ * What sits behind the globe.
+ *
+ * MapLibre's atmosphere leaves everything past its own glow transparent, so
+ * the page shows through around the sphere. A flat map covers the canvas edge
+ * to edge, which is why this is only ever visible under the globe projection.
+ * The colours are the basemap's own land and night sky rather than a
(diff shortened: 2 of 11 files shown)


alexwohlbruck:

PR #438, opened 2026-09-02
Interchange lines

alexwohlbruck:
Tapping the merged label a map draws over an interchange opened the whole
group's departures — #437 — but the bullets beside the station's name still
came from one platform group. Brooklyn Bridge–City Hall read `4 5 6 6X` with
the J and Z filed under Transfers, which is the answer for tapping one station,
not the group.

The badge row is built from `/transit/routes`, and that call was made with
neither `complex` nor the place's coordinates. Both are now sent:

* **`complex`** — passed through from the tap, so a merged label asks about the
  whole interchange. Needs barrelman alexwohlbruck/barrelman#89, which adds the
  parameter to `/transit/routes`; without it the response is unchanged.
* **`lat`/`lng`** — which is what turns on nearby connections. Barrelman has
  reported `via: 'nearby'` since 0.2.6 and nothing ever asked for it, because
  the point it needs was never sent.

Sending the second exposed a latent bug worth naming: the badge row selected
its lines with `via !== 'transfer'`, which stopped meaning "the station's own"
the moment a third kind existed — every nearby bus would have landed under the
station's name. It now selects `via === 'station'` positively, and nearby lines
join the Transfers section after the published ones, nearest first.

## Tests

`bun run test` in `web/` — 1719 pass. Server typecheck is failing across ~70
files on `dev` already; no new errors in anything touched here.

Preview: https://vega5.tailc47d6.ts.net:8404


alexwohlbruck:

alexwohlbruck:
Verified end to end against vega2 now that barrelman 0.2.7 is deployed there — this is the first time the whole chain could actually run.

Brooklyn Bridge–City Hall (`/place/node/8410411844?tab=transit`):

| tap | header bullets | Transfers section |
|---|---|---|
| merged label (`&complex=1`) | **J Z 4 5 6 6X** | M22 |
| single station | **4 5 6 6X** | J, Z, then M22 |

Before this branch the merged tap gave `4 5 6 6X` with J and Z under Transfers — the single-station answer for a group tap.

The M22 (Lower East Side–Battery Park City) is `via: 'nearby'` appearing for the first time. Barrelman has reported it since 0.2.6 and nothing ever asked, because the coordinates it needs were never sent. Published transfers still lead, the walked-to bus follows.

Two things this run turned up:

- **The "Reachable from this station without exiting" subtitle was still there.** It was supposed to have come out earlier and didn't — `StationTransfers.vue` still rendered `place.transit.transfersHint`. Removed in 405b371, along with the now-unused key in `en-US` and `es-ES`.
- **The preview server does not hot-reload.** `parchment-preview-server@.service` runs `bun src/index.ts` with no `--hot`, so server changes need `systemctl --user restart parchment-preview-server@<slot>`. The web side does reload. Worth knowing — the header looked unfixed until I restarted it, and the code had been right the whole time.

No screenshot: the browser automation's snapshot fails against this page (the map pegs the main thread), so the table above is read out of the live DOM instead of a picture of it.

PR #439, opened 2026-09-02
Canvas order and speed

alexwohlbruck:
A review pass over the canvas feature — the renderer, the drawing tools, the editor, the store and the API. Three of the things it turned up are bugs; the rest is what a canvas costs to edit.

## Bugs

**Reordering the stack told the map nothing.** A render pass skipped any layer whose configuration was unchanged, and nothing in a configuration records where a layer draws — so two swapped layers both looked untouched and the drag reached the engine as no calls at all. The order only took effect when something else forced a redraw. The same blind spot ran the other way: the engine has no "move this layer", so re-adding a restyled one quietly lifted it above everything it had been under.

A pass now keeps the longest run of layers that are unchanged *and* already in the right place, and lays the rest down again in order above it. Visibility is tracked apart from the configuration, so flipping a switch is one call rather than a rebuild that also reorders.

**Reshaping the mark at the foot of a run tore the run off the map.** Marks that sit together share a source and a set of style layers, named after the mark at the bottom — but the name was taken after the suppressed and hidden marks had been filtered out, so picking that mark up renamed every layer in the run: 8 removals and 8 additions, twice per drag. It takes its name from the stack now, before anything is held out.

**Changing a travel-time area's reach while the engine was thinking lost the mark.** The abandoned request's error handler cleared the origin, and the answer that was still on its way arrived to find nothing to attach itself to. `useLatestRequest` now distinguishes a request that failed from one a newer ask replaced.

## Speed

A canvas can hold 4 MB of imported GeoJSON per layer (`MAX_IMPORT_BYTES`), and every edit was writing the document out three or four times over: once for the dirty check, twice to take an undo step, once more to work out what the map had to be told — plus a deep watch walking every feature to notice the edit at all. Measured on the dev box: 21 ms to serialise a 4 MB body, 60 ms to parse one back, 36 ms to walk one. So roughly 80 ms an edit before anything was saved, and more on the save tick; several times that on a phone.

Nothing is copied to answer any of those now:

- **Dirty** is the body compared to the one that reached the server, by identity — every edit replaces the body rather than changing it in place. Undoing back to what was saved reads as saved, because the step put the very same body back.
- **Undo steps** hold the editor's own values instead of a JSON string, so a hundred of them share everything the edits did not touch. The stack held up to 100 full copies before.
- **Both watchers** see a replacement without walking what is inside it.
- **A source** is compared as two halves, its shape and its data, and the data half is remembered against the object it came from — so an unchanged inline dataset is recognised by identity rather than serialised and parsed on every pass.
- **A run of marks** asks for the layers it uses rather than every layer any mark could need. A run of plain lines is one layer now instead of eight — a fill, four sets of pin layers, a halo and a label layer that could never match a feature.
- **A save** no longer gets the whole document echoed back at it. `PUT /library/canvases/:id` answers without the body; both callers overwrote it with their own copy the moment it arrived.

## Tidying

One place mints an id for something on a canvas, rather than the same expression in five files. One table holds the range each numeric setting is offered over, rather than the tool options bar and a mark's own editor each carrying their own idea of how thick a stroke can get.

Two new composables replace duplicated machinery rather than adding any: `useDrawingSurface` is the shape of the engine's map instance that drawing needs, written once instead of nine one-off casts; `useLatestRequest` is the abort-and-supersede dance that four places were each doing their own way. `useCanvasAnnotations` loses 119 lines and `useAnnotationEditing` 49, net −92 across the two with both helpers included.

## Verified

`vue-tsc` clean, **2129 web tests** (18 new), **3942 server unit** and **73 DB-backed** all pass. The isochrone test fails against the old code; so does each of the reorder tests.

Also driven in a browser against the running preview, since the reconciliation is the sort of thing a unit test can agree with and a build still get wrong:

```
initial       addLayer:cl-a, addLayer:cl-b
reorder       addLayer:cl-b, addLayer:cl-a      (was: nothing at all)
hide cl-a     visibility:cl-a:false             (was: a full rebuild, and a reorder with it)
```

No screenshot: every change here is either invisible in a still frame (draw order, a run that no longer flashes) or a number rather than a picture, and the preview browser is signed out so there is no canvas to photograph. The preview is up if you want to drag a few layers around on a phone.

## Left alone, deliberately

- The shared-link view lists `body.layers` flat, ignoring groups, order and marks — so it disagrees with the owner's panel and with what is drawn. Fixing it means teaching the group and mark rows a read-only mode, which is a UI change rather than a cleanup, so it wants its own PR.
- `GET /library/canvases` returns every body, so opening the Canvases tab fetches, decrypts and localStorage-writes every canvas in full to render names and counts. The bodies of canvases switched on are genuinely needed at boot, so splitting this is a design decision, not a tidy-up.
- `parchment-docs` still has no canvases documentation at all — nothing in `content/` mentions them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 * Pins on a canvas are drawn the way search results are, glyph and all — the icon you choose is the icon on the map, at the same size whatever the zoom. Saved places shrink and fade as you zoom out because the question there is "where have I saved things"; a pin you placed deliberately stays where and what you put it
 * Canvas layers keep their colour on a night basemap. Mapbox dims the map after dark, which is right for buildings and wrong for something you drew, so marks and data layers now light themselves
 * A canvas's layers and marks are one reorderable list now, bottom to top, so what covers what is whatever the list says. Related things can be gathered into a group with its own name and switch
+* Canvases holding large imported datasets stay responsive as you edit them, and save without sending the whole document back and forth
 * A pin on a canvas can be a circle, a square or just its icon. A square reads as a station or a stop the way the basemap already draws them, and a bare icon keeps a busy map quiet
 * Canvases appear in the layer selector under their own group, so one can be switched on or off from the map without opening your library
 
@@ -51,6 +52,10 @@
 * The tick beside a custom colour now applies it and closes the picker, rather than leaving the picker open with nothing apparently happening
 * Escape works again in views that bind it alongside another. Several parts of the app listen for Escape at once, and closing any one of them quietly took the others' handling with it — so Escape could stop dismissing a panel until the page was reloaded
 * Undo and redo on a canvas now work while the cursor is in a text field. Naming a mark and then pressing ⌘Z used to do nothing, though the toolbar button worked
+* Reordering a canvas's layers and marks now changes what covers what on the map straight away, instead of only after something else made it redraw
+* Restyling a canvas layer no longer quietly lifts it above the layers it was sitting under
+* Marks around the one you are reshaping or hiding no longer flicker off the map and back
+* A travel time area is no longer lost when you change its mode or reach while it is still being worked out
 
 ## [0.8.3] - 2026-09-02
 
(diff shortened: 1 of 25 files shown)


alexwohlbruck:

PR #443, opened 2026-09-02
Layer selector tabs

alexwohlbruck:
## Summary

- separate map overlays, saved-place collections, and canvases with the app’s existing linear tabs
- tighten layer rows and base-map cards with the standard app radius
- preserve real collection and canvas controls, including group visibility and empty states
- add mobile sheet clearance so tabs do not collide with the close control

## Validation

- ./node_modules/.bin/vue-tsc --noEmit
- ./node_modules/.bin/vitest run src/lib/saved-places-layers.test.ts src/lib/canvas-layers.test.ts (32 tests)

## Preview

https://vega5.tailc47d6.ts.net:8402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 
 ### Changed
 
+* The layer selector now separates map overlays, saved-place collections and canvases into a compact tabbed panel with a clearer base-map picker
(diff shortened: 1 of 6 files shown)


alexwohlbruck:

PR #444, opened 2026-09-02
Keep rails under transit layer

alexwohlbruck:
Turning on a transit layer erased the railway network from the basemap.

## Cause

`setTransitLabels(false)` fires whenever a transit layer becomes visible, to stop the basemap's own stop labels doubling up under the overlay's stops. On MapLibre that hides `layerGroups.transit`, which was derived by matching layer **ids**:

```ts
transit: idsWhere(l => /rail|transit|subway|tram|funicular/i.test(l.id)),
```

In the current spec the only ids containing those words are the six railway *line* layers — `Major rail`, `Minor rail`, `Railway tunnel` and their hatchings. So the toggle hid exactly the wrong thing:

* it erased the rail network geometry, and
* it hid **zero** labels, so the basemap's stop labels stayed doubled under the overlay — the thing it exists to prevent, and the behaviour `transit-poi.mjs` documents.

Mapbox was unaffected; its strategy delegates to the `showTransitLabels` basemap config, which already does the right thing. MapLibre-only.

This also explains the **Transit labels** setting in Appearance behaving oddly — it was toggling rail geometry rather than labels — and the same setting now carried per-canvas.

## Fix

Derive the group from what the layers actually draw rather than what they're called: POI symbol layers whose `class` filter is entirely transit classes. That resolves to `Station` (bus/railway stops).

Kept derived from the spec rather than hand-listing `'Station'`, matching the existing intent of `idsWhere` — "so a regenerated spec cannot silently drop one out of a toggle".

Ferry and harbour stops live in the mixed `Transport` layer alongside parking and fuel, so they can't be hidden by layer visibility without taking unrelated POIs with them. Left alone rather than widening the change.

## Testing

* New regression test: the transit group is only POI symbol layers, and never a line layer.
* `map-style.test.ts` — 155 passing.
* `src/services/layers` — 148 passing.
* `vue-tsc` clean.

No screenshot: the headless browser can't capture this app since the WebGL globe landed — snapshots fail and the tab hangs. Worth confirming by eye on the preview.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 
 ### Fixed
 
+* Turning on the transit layer no longer erases the railway network from the map. The basemap's station labels step aside for the layer's own stops instead, so the rails stay drawn
 * Tapping a station label that covers a whole interchange now opens all of it. New York draws four separate stations named "Canal St" as one symbol, and tapping it opened whichever of the four the label happened to sit on — so the panel showed two lines where the map had drawn six. Tapping a single line's marker still opens just that station
 * The map keeps up with the sidebar as it is dragged or collapsed, resizing frame by frame rather than snapping once the panel has settled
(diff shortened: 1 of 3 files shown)


alexwohlbruck:

PR #448, opened 2026-09-04
Compact routing popup

alexwohlbruck:
## Problem

Opening the routing settings popover on directions with a short window pushed most of its content — Transit Options, Advanced, Done — off the bottom of the screen with no way to reach it. The layout was also looser than it needed to be.

## Changes

**Scroll instead of overflow** — the desktop popover in `ResponsivePopover.vue` had no height constraint. It now caps at reka-ui's `--reka-popover-content-available-height` (the space between the trigger and the viewport edge) and scrolls internally. Applies to every popover built on `ResponsivePopover`; the mobile bottom-sheet path already handles its own scrolling and is untouched.

**Compact layout** — the routing engine picker is hidden when only one engine is configured (it's still applied silently as the default); section padding and row spacing are tightened throughout; and Max transfers is a label + narrow input row instead of a full-width field.

## Verification

Checked live in a 1100×660 window: the popover caps at exactly the available height and scrolls cleanly to the Advanced section and Done button, which were unreachable before. With the engine list reduced to one, the Routing Engine row disappears and General Options leads straight into the toggles; content height dropped from 559px to 477px.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* The routing settings popup no longer overflows off screen — when its options are taller than the window, it caps at the visible height and scrolls inside.
+
 ## [0.9.0] - 2026-09-03
 
 ### Added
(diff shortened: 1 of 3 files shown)


alexwohlbruck:

PR #449, opened 2026-09-04
Transit lines in search

alexwohlbruck:
Search now surfaces barrelman's transit results (shipped in barrelman 0.2.15/0.2.16, live on production): GTFS lines and GTFS-only stops, opened from the palette and the results page into the right views.

## How it works

Barrelman marks transit hits with `kind: 'transit_route' | 'transit_stop'` and a `transit` object carrying the `feedId`/`routeId`/`stopId` the `/transit` endpoints are keyed by.

**Server** — the barrelman integration adapts these into pseudo-`Place`s carrying `transitLine`/`transitStop` refs, so they ride the existing search pipeline (merge, scoring, interleave) untouched; the search service then types them `transit_route`/`transit_stop` in both the full and autocomplete responses.

**Web**
- A line opens the existing transit route detail (`/transit/route/:feedId/:routeId`). `getPlaceRoute` learned the `transit-route/<feedId>:<routeId>` id form, which makes the palette's regular place path and PlaceCard routing both work unchanged.
- **Lines wear their route bullet in the palette** — the short name on the line's GTFS colour via the shared `RouteBullet` (so a wordy short name becomes a pill, same as departure boards), with the long name as the row title.
- A GTFS-only stop re-checks portolan's stop→OSM index client-side, opening the OSM station (`?complex=1`) when matched, else the name+coords place view with the transit widget expanded.
- The full results page includes the new types in its place list.

## Verified live on the preview

- "7" (single character) → the Flushing Local and two other 7s, each with its coloured bullet; clicking opens `/transit/route/5/7`.
- "q" → Q · Broadway Express on MTA yellow.
- "franklin ave medgar" → Franklin Avenue-Medgar Evers College first (barrelman's abbreviation expansion, 0.2.16).
- `vue-tsc` clean; all web tests pass; server search suite passes.

## Follow-up

Digit-leading queries ("82 st jackson heights") still rank Pelias addresses above a full-name station match — the station appears, just below the addresses. Loosening that would touch genuine address-search ranking, so it's deliberately left for its own change.


alexwohlbruck:

PR #450, opened 2026-09-04
Trim unsupported directions modes

alexwohlbruck:
Removes the **wheelchair travel mode** — the selectable mode/profile — from directions and the router.

## Scope

**Server** — `TravelMode.WHEELCHAIR`; `'wheelchair'` from `VehicleType` / `Mode` / `SelectedMode`; `planWheelchairSegment()` and its `getModesToGenerate` case; the mode's CO2 and safety-score entries; the ORS `wheelchair` profile literal; provider `supportedModes` and profile mappings (GraphHopper, Valhalla, Barrelman, Dawarich); the GraphHopper wheelchair custom-model rules; the API schema literals and the `/directions/vehicle-types` entry.

**Web** — the mode's type unions; the wheelchair preferences tab (tab grid 5→4) and its store defaults; mode icons; trip-signature token; vehicle-mode mapping; route-profile tabs; segment/trip colour and chart-eligibility checks; the Directions mode selector entry; en-US/es-ES strings. Unused `AccessibilityIcon` imports went with them.

**Docs** — `server/docs/trip-planner.md` updated. `docs/trip-planner-review.md` is a marked historical snapshot, so this adds to its existing "since then" banner rather than rewriting the record.

## Kept deliberately

Three things share the word but are not the mode:

- **`wheelchairAccessible`** — the routing *preference* under Walking and Transit (avoids stairs, prefers accessible entrances). Independent of the mode; still drives Valhalla's pedestrian type and station-entrance snapping in `trip.service.ts`.
- **Station/stop accessibility** — `wheelchairBoarding`, entrance `wheelchair` flags from GTFS/Barrelman.
- **OSM place tags** — `wheelchair=yes` chips, place details, tag labels/icons.

## Notes

- One web test asserted that `localStorage` round-trips `selectedMode: 'car'` — never a valid mode. Since a persisted `'wheelchair'` would now be sent to an API that rejects it, `loadSelectedMode()` gained a validation whitelist, and that test was replaced with a valid-mode case plus a fallback case.
- `bun run openapi` also picked up six stale `/subscriptions/*` paths missing from the committed spec. The file is fully generated, so they are kept rather than hand-trimmed out of the diff.

## Verification

- Server: 3935 tests pass, 0 failures. Typecheck adds no new errors (the baseline is noisy; this diff only removes errors).
- Web: 2146 tests pass, 0 failures. `vue-tsc` clean apart from one pre-existing unrelated error (`@tmcw/togeojson` missing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Removed the wheelchair travel mode from directions. Accessible routing stays available as the "Wheelchair accessible" preference under Walking and Transit, which avoids stairs and prefers accessible stops and entrances.
+
 ### Fixed
 
 ## [0.9.0] - 2026-09-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Removed the wheelchair travel mode from directions. Accessible routing stays available as the "Wheelchair accessible" preference under Walking and Transit, which avoids stairs and prefers accessible stops and entrances.
+* Hid the rideshare travel mode until there is a working rideshare integration behind it.
 
 ### Fixed
 
--- a/docs/trip-planner-review.md
+++ b/docs/trip-planner-review.md
@@ -7,9 +7,12 @@
 > the rideshare `isAvailable()` typo (C1) is fixed and rideshare+transit variants are
 > now derived from intermodal results; vehicle-access queries gained origin→vehicle
 > walk legs, `useKnownVehicleLocations`/`maxWalkingDistance` handling, and
-> `parkedVehicles` tracking. Line numbers and the two-path architecture described
-> below no longer match the code — read this for the *why* behind the consolidation,
-> not as a map of the current source.
+> `parkedVehicles` tracking; and the wheelchair travel mode was **removed** —
+> `planWheelchairSegment` and the `wheelchair` mode/profile no longer exist (the
+> `wheelchairAccessible` routing preference for walking and transit remains).
+> Line numbers and the two-path architecture described below no longer match the
+> code — read this for the *why* behind the consolidation, not as a map of the
+> current source.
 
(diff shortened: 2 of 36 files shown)


alexwohlbruck:

alexwohlbruck:
Verified in a running preview of this branch:

- **Mode selector** — All modes, Walking, Cycling, Transit, Driving, Rideshare. No wheelchair entry.
- **Route preferences** — the mode tab row is now `grid-cols-4` (Walking / Cycling / Transit / Driving) and lays out evenly; no wheelchair tab, no gap where it used to be.
- **Stale persisted mode** — with `localStorage.selectedMode = 'wheelchair'` set by hand, Directions loads normally and falls back to `multi` instead of sending a mode the API would now reject.
- No occurrence of "wheelchair" anywhere in the rendered Directions UI.

Note: on the default GraphHopper/Barrelman engine the "Wheelchair accessible" walking toggle stays hidden, because that engine declares the preference unsupported (`wheelchairAccessible: false`). That gating is pre-existing and unchanged — the toggle still appears on engines that support it, e.g. Valhalla.

Test suites: server 3935 passing, web 2146 passing.

alexwohlbruck:
Added a second commit: **hide the rideshare travel mode** (`3ccc957b`).

Rideshare has no working integration yet, so the mode is hidden rather than removed — this is meant to be reverted when one lands.

- `RIDESHARE_ENABLED = false` in `directions.store.ts` is the single switch. Flipping it to `true` restores the mode in both places that consume it; the mode entry, icon, label, types and trip-rendering paths are all left intact.
- The Directions mode selector filters the entry out.
- `SELECTABLE_MODES` (hoisted to module scope so both files share one source of truth) drops it too, so a user who had rideshare selected falls back to `multi` instead of sitting on a mode with no visible tab.
- Added a store test covering that fallback.

Scope note: the server already only generates rideshare trips when a rideshare integration is configured (`rideshareService.isRideshareAvailable()`), so nothing else was needed for the current state. If someone did configure Uber/Lyft, All-modes results could still surface a rideshare trip while the mode itself stays hidden — suppressing that would be a server change, and I left it alone as beyond "hide it for now".

Web suite: 2147 passing. `vue-tsc` clean.

alexwohlbruck:

PR #451, opened 2026-09-04
Mobile canvas toolbar

alexwohlbruck:
Mobile fixes for the canvas editor.

**Toolbar floated on top of the expanded bottom sheet.** The toolbar's wrapper was `z-40` — the same z-index as the mobile content sheet — and, being teleported to `body` later, always won the tie. It now sits at `z-[35]`: above the sheet's overlay (30) and the nav sheet (30), below the sheet content (40), so dragging the sheet up slides it over the toolbar.

**Tools wrapped at an arbitrary point on narrow screens.** The strip was one `flex-wrap` container, so the break landed wherever width ran out. It is now two non-wrapping units inside the wrapping container, so the break always falls at the same boundary: select + the eight drawing tools on one row, and the edit actions — erase, destination picker, undo, redo/done — below. Erase moved from the tools array into the actions group, since it removes marks the way undo does. The divider between the two groups hides below `sm` so a wrapped row doesn't start with a dangling separator.

**Tools overflowed the clipped box below ~350px.** Nine 32px buttons don't fit a narrow phone, and the box is `overflow-hidden`, so the end tools were simply lost. Below 380px the strip steps down to 28px buttons with 14px glyphs — the grouping survives and every tool stays reachable. Desktop is unchanged.

**Drawing popped the sheet open after every mark.** Committing a mark opens its row with the name field focused; on a phone that focus rides in with the keyboard and drags the sheet open mid-draw. The autofocus is now desktop-only — the row still opens, and a tap on the field starts naming.

**Minimized, the sheet showed an arbitrary 160px slice.** The editor registers its header as the sheet's dynamic-peek region (new opt-in `peek-header` prop on `DetailPanelLayout`), so the collapsed sheet rests exactly around the canvas icon, name, description and nav buttons.

**The middle detent sat too high.** It was 60% of the screen, which buries the map you're drawing on; it now rests at 40%.

Verified in the running app. At 320px: 28px buttons, first and last tool both inside the box, no overflow, wrap exactly between draw tools and actions. At 390px: 32px buttons, same wrap, nothing clipped. With the sheet expanded a hit-test over the toolbar lands on the sheet; minimized (78px peek hugging the header) placing pins leaves the sheet at peek with focus untouched; the middle detent measures 40%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* A minimized bottom sheet stays put when you tap the map behind it, instead of springing back to half height — most visible when drawing on a canvas from a phone.
 * Drawing on a canvas from a phone no longer pops the panel open after every mark, and the minimized panel shrinks to just the canvas's title bar.
 * On a phone, the canvas toolbar no longer floats on top of the expanded panel — the panel now slides over it. When the toolbar is too wide for the screen, it breaks into a fixed pair of rows: drawing tools above, erase and undo/redo below, stepping down a size on the narrowest phones so no tool is cut off.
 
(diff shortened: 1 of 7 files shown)


alexwohlbruck:

PR #453, opened 2026-09-04
In-app feedback form

alexwohlbruck:
Replaces the Feedback link (which pointed at GitHub issues) with a proper in-app feedback form, backed by a self-hosted [Quackback](https://github.com/QuackbackIO/quackback) instance.

Quackback was picked because it is the only actively-maintained open-source feedback platform with a real two-way Linear integration — OAuth, issue creation, inbound webhooks and status mapping. Closing a synced Linear issue flips the linked post and notifies everyone who voted for it.

## What's here

**Server** — a `feedback` module proxying Quackback's REST API:
- `GET /feedback/boards` lists the boards feedback can be filed against (cached 5 min)
- `POST /feedback` mirrors the Parchment user into Quackback via `/users/identify`, then files the post attributed to them, so they get notified when it ships
- Rate limited to 5 submissions per minute per user

**Web** — a `FeedbackDialog` styled to the app: board pills, summary, details. The `useFeedback` composable probes once per session and doubles as the board prefetch.

**Shake to report (mobile)** — shaking the phone opens the form, so reporting something you just hit takes no hunting through menus. `useShakeGesture` requires three hard jolts inside a second before firing, with a cooldown, because a false positive (a dialog over the map while walking) costs far more than shaking again. Off by default and opt-in under Settings → Behavior — flipping that switch is also the user gesture iOS requires before it will grant motion access, and a denial flips it back rather than leaving the setting reading as on while nothing listens.

## Configuration

There are no environment variables. An admin connects the instance under **Settings → Integrations → Quackback** (instance URL + API key), matching how the OpenStreetMap OAuth app is configured — a system-scoped, config-only integration. Leave it unconfigured and both feedback entry points hide themselves, so self-hosters don't get a broken button.

The key must be an **admin-role** Quackback key — attributing a post to its actual author is admin-gated upstream, and a member-role key would file everything under the key's own account.

Linear is connected on the Quackback side. A self-hosted Quackback stores its own OAuth-app credentials in the database via its admin settings UI (`DbCredentialSource`); the `LINEAR_CLIENT_ID` env vars are the managed-cloud path and don't apply.

## Note on licensing

Quackback ships an `@quackback/widget` npm package, which would have been the quick path. It is AGPL-3.0, and bundling it would pull Parchment's frontend into AGPL copyleft — incompatible with our Apache-2.0 + Commons Clause. Hence the native dialog against the REST API. Self-hosting the Quackback server is unaffected; it's a separate service.

## Tests

11 server tests (board mapping, caching, author attribution, URL building, unconfigured/half-configured states, error surfacing) and 11 web tests (probe-once, refresh, unavailable states; shake threshold, single-knock rejection, gentle-movement rejection, double-fire suppression, cooldown, teardown).

Full web suite green (2156 tests). Full server suite shows no new failures against the `dev` baseline.


alexwohlbruck:

alexwohlbruck:
Verified end-to-end against a Quackback instance wired up to the slot 8 preview.

**Boards** — `GET /feedback/boards` returns the two boards created upstream (Bugs, Feature requests); they render as pills and the selected board's description shows beneath them.

**Submission** — filed one post per board from the dialog. Both landed on the correct board and, importantly, came back attributed to `Alex Wohlbruck` rather than to the API key's own principal, which confirms the `/users/identify` → `authorPrincipalId` path works and that the key needs the admin role:

```
post_01m1nf2pc5fk0t495yf43gbyp1 | Bugs             | Transfers list is cut off on iPhone SE
post_01m1nf3qaafk0t497feytfs7gk | Feature requests | Offline map packs for a whole metro area
```

(Both test posts deleted afterwards.)

**UI** — dialog closes on success and the toast reads "Thanks for the feedback / We'll let you know when there's an update." with a working **View** action. Unauthenticated `GET /feedback/boards` returns 401.

**Fix along the way** — `ResponsiveDialog` exposes a named `content` slot rather than a default one, so the first pass rendered only the dialog header. Corrected in e08aeed7.

Tests: 9 server + 5 web, all passing. Full server suite shows no new failures against the `dev` baseline (589 failures on this branch vs 591 on `dev`, all pre-existing from the test database not running locally).

alexwohlbruck:
Follow-up: moved the Quackback connection off environment variables onto a system integration, and added shake-to-report.

**Correction to my earlier note on Linear.** I said to set `LINEAR_CLIENT_ID` / `LINEAR_CLIENT_SECRET` in Quackback's `.env`. That is the managed-cloud path. Self-hosted Quackback defaults to `DbCredentialSource`, which reads `integration_platform_credentials` — "self-hosters own their own OAuth apps and paste their own credentials" via the admin settings UI. So Linear was always admin-configured; no env editing needed.

**What changed here.** `QUACKBACK_URL` / `QUACKBACK_API_KEY` are gone. `IntegrationId.QUACKBACK` is a system-scoped, config-only integration following the `OpenStreetMapSystemIntegration` pattern, resolved per request so an admin can change it without a restart.

Verified the integration class against the live instance:

```
validate ok  → true          validate bad → false
test good    → Connected — 2 boards available
test badkey  → Invalid API key
test badhost → ECONNREFUSED
test empty   → Instance URL and API key are required
```

**Not verified.** The browser automation host died partway through, so I could not re-check the Settings → Integrations form rendering or exercise the shake gesture on a real device. The service-layer read path is covered by unit tests and the catalog entry mirrors the OSM one field-for-field, but the form rendering and a real shake are unconfirmed — worth a look on the preview before merge.

Tests: 11 server + 11 web. Full web suite green (2156). Server suite unchanged against baseline.

alexwohlbruck:
Preview respun on slot 1 — https://vega5.tailc47d6.ts.net:8401 — and the two things I could not confirm last time are now verified, plus one bug that respin caught.

**Bug found and fixed.** The integration card rendered its description as a raw i18n key (`settings.integrations.descriptions.quackback`) — the catalog's `description` field is not what the UI shows; it looks the string up in `settings.integrations.descriptions.<id>`. Added for both locales in a45dd78b. Only visible once the card actually rendered, which is why the earlier unit-level checks missed it.

**Settings → Integrations.** Card renders, config form shows both fields, **Test Connection** returns *Connection successful* against the live instance, and Save persists a system-scoped row:

```
quackback | system_scoped=t | 2026-09-04 15:09:51
```

**Feedback appears once configured.** The preview DB is a fresh clone of `dev`, so feedback started hidden — exactly the intended unconfigured behaviour. After saving the integration the nav item appeared and the dialog loaded both boards from it. No restart needed, confirming the per-request resolution.

**Shake to report.** At iPhone 12 Pro width the toggle renders under Settings → Behavior, flips to `aria-checked=true` and persists (`shake-for-feedback=true`). Firing synthetic `devicemotion` events against the running app:

- gentle wobble (±2, 8 samples) → dialog does **not** open
- real shake (±30, 6 samples) → feedback dialog opens with boards loaded

Still unconfirmed: the iOS `requestPermission` branch. Chromium has no permission gate, so the preview exercises `canStartWithoutPrompt`; the iOS path needs a real device.

Shake is left enabled on the preview if you want to try it from a phone.

alexwohlbruck:
Feedback is now gated on a capability rather than on the Quackback integration by name.

**Server.** New `IntegrationCapabilityId.FEEDBACK` and a `FeedbackCapability` (`listBoards`, `submitFeedback`). The Quackback integration implements it — the REST calls, board cache and author attribution moved out of `feedback.service.ts` and into the integration, matching how OpenWeatherMap owns `getWeather`. `feedback.service.ts` is now a thin resolver over `getConfiguredIntegrationsByCapability(FEEDBACK)`, so a second provider needs no change to it. `config/feedback.config.ts` is gone.

**Client.** `isFeedbackActive` in the integrations store mirrors `isRoutingActive` — any configured integration with the capability active. `useFeedback.available` reads it, so gating is synchronous and the startup probe to `/feedback/boards` is deleted; boards are fetched only when the form opens.

**Two bugs this surfaced.**
1. My catalog entry hardcoded `capabilities: []` instead of the `get capabilities()` accessor every other integration uses. `createIntegration` derives a row's capabilities from that, so newly saved rows would have stored `[]` and feedback would never have activated. Now derived from the manager.
2. Capabilities are persisted per row and read back from it, not re-derived from the definition — so the row I saved before this change carried `[]`. Only affects rows written against the old definition; nothing in production has one, since this is unreleased. Worth knowing if you re-test on an existing preview: delete and re-add the integration.

**Verified on the preview** (https://vega5.tailc47d6.ts.net:8401), all four states:

| State | Result |
|---|---|
| Integration absent | Feedback hidden from nav |
| Added via UI | Row stores `[{"id":"feedback","active":true}]`; nav shows Feedback; dialog loads both boards |
| Capability set `active:false` | Feedback hidden — the case you asked for |
| Re-enabled | Feedback returns |

The card also renders a **Feedback** capability chip and appears under the Feedback capability filter.

Tests: 26 server (16 integration, 10 service resolution) + 6 web. Full web suite green (2157); server suite unchanged against baseline.

alexwohlbruck:
Now pointed at the live instance — https://feedback.parchment.app — and adding a second board surfaced a real bug.

**Bug: pressing the selected board pill killed the form.** A single-select `ToggleGroup` deselects when you press the active item, so tapping the already-chosen board left `boardId` empty, disabled Submit, and showed no error — a dead form with no explanation. It only shows up with more than one board, which is why the earlier single-board testing missed it.

Fixed by ignoring the empty emit rather than using `v-model`. This repo already documents the same reka behaviour in `ToggleGroup.test.ts` ("including null when the active item is deselected"), so the guard matches the established convention.

```
before → Feature Requests:off  Feedback:off   submitDisabled: true
after  → Feature Requests:on   Feedback:off   submits fine
```

Added `FeedbackDialog.test.ts` (5 tests: preselect, re-press keeps selection, switching, submits against the selected board, no submit without a title). Confirmed it's a real regression test — reverting the one-line fix fails exactly the re-press case and nothing else.

**Live wiring verified.** Admin-role key minted on the production instance, integration saved, Test Connection successful. Submitted from the app to each board: both landed on the right one attributed to `Alex Wohlbruck`. Both test posts deleted; the board is empty.

Also copied SMTP across from the `personal-site` container (Gmail, port 465) so shipped-post notifications actually send. Two adjustments were needed: `EMAIL_SMTP_SECURE=true`, which is absent from Quackback's `.env.prod.example` but is the only thing that turns on implicit TLS — without it nodemailer attempts STARTTLS on a 465 socket; and `EMAIL_FROM` set to the Gmail address, since Gmail refuses to send as an unverified alias like `noreply@parchment.app`. Verified with a real SMTP_SSL login: AUTH OK.

Full web suite green (2162, up 5).

PR #455, opened 2026-09-04
Rider-friendly transit subtitles

alexwohlbruck:
Search subtitles for transit hits now read the way a rider talks, built from route_type (basic + extended GTFS codes) with agency as tiebreaker — surveyed against every route_type actually present in production:

- "MTA New York City Transit · Subway line" (was "… · subway")
- "Commuter rail line" for LIRR, whose feed names no agency (was the bare word "rail")
- "Amtrak · Intercity rail line" vs "Metro-North Railroad · Commuter rail line" — same GTFS code 2, different promises
- "Shuttle bus route" for the MTA's 260 extended-code-711 routes; buses run *routes*, rail runs *lines*
- Stops get the same treatment: "Subway station", "Bus stop", "Ferry terminal" instead of "Transit stop"

Server-only (label engine in `server/src/lib/transit-mode-label.ts`, unit-tested); no barrelman change needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Fixed
 
+* Transit search results describe themselves the way a rider would — "MTA New York City Transit · Subway line", "Commuter rail line", "Shuttle bus route" — instead of a bare lowercase mode word
+
 ## [0.10.0] - 2026-09-04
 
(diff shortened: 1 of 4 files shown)


alexwohlbruck:

PR #460, opened 2026-09-06
Show transit through buildings

alexwohlbruck:
Transit ribbons stopped reading through the buildings they pass behind: on the tilted map a route simply vanished into a block.

MapLibre has no depth-aware line occlusion, so the overlay draws each ribbon twice — the solid line below the basemap's extrusions, a dimmed copy above them. The dimmed copy is inserted before the basemap's first label layer, and in the new style that layer is `Oneway`, which sits *below* `Building 3D`. Both copies ended up under the towers.

The ghost now anchors on the first label layer drawn above the buildings. Mapbox is untouched — it uses `line-occlusion-opacity` and was never affected.

Also fixes a duplicate `addSymbolLayers()` call from the same commit: the second one threw "layer already exists" and aborted the rest of that sync pass (stations, tile filters, hydration).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+* Transit lines show faintly through the buildings they pass behind again, instead of disappearing into the block.
+
 ## [0.10.1] - 2026-09-04
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


alexwohlbruck:

PR #466, opened 2026-09-06
Pin sheet footers

alexwohlbruck:
The mobile dashboard's recents list is endless, so the account dropdown parked
below it was unreachable — you had to scroll past every recent place to get to
your own account. It's now pinned to the bottom of the sheet, with the list
scrolling between the search field and it.

Getting there meant writing down how scrolling inside a bottom sheet actually
works, and then fixing the views that were doing it wrong.

## The layout contract

Every sheet is now three optional regions, documented at the top of
`BottomSheet.vue` and in [parchment-docs](https://github.com/alexwohlbruck/parchment-docs/pull/9):

```
sticky header    <SheetHeader>   pins to the top; sizes the peek detent
scrollable content               the one and only scroll surface
pinned footer    <SheetFooter>   rides the sheet's visible bottom edge
```

Both new components live in `components/sheet` and render in place, so they work
the same from a routed view inside the mobile sheet, from a direct child, and
inside the desktop side panel.

**Why the footer needs a portal.** A snap-point sheet is a full-viewport-tall
panel translated down by the detent, so its own bottom edge sits far below the
fold — `position: sticky; bottom: 0` sticks to an off-screen edge. The sheet now
keeps a thin layer pinned to its *visible* bottom edge (sized by a new
`--sheet-visible-height`) and `<SheetFooter>` teleports into it. The layer holds
only the footer, so resizing it during a drag never touches the content's
layout. The peek detent grows by the footer's height, so a collapsed sheet shows
its peek content *and* its footer.

## Views fixed

The shape to look for is `h-full` + an inner `flex-1 overflow-y-auto`: the inner
scroller swallows the upward pan, so the sheet can't be expanded from inside its
content or dragged back down once scrolled. Removed from:

- `DetailPanelLayout` (collections, notes, layer and canvas editors, public canvases)
- Search results — infinite scroll now listens to the sheet's surface via `findScrollAncestor`, matching what the dashboard already did
- Timeline, `BookmarkList`, friend and tracker details
- Library index views, which were `h-full` where they wanted `min-h-full`

The note detail's comment box became a real `<SheetFooter>` — it was only
pinning by accident, via the scroller this removes.

## Checks

- `vue-tsc` clean, 2183 unit tests pass
- Verified on the preview at iPhone 12 Pro size: footer sits on the viewport bottom at the peek detent, stays there through a drag to full height, and content scrolls clear of it. Collection detail's header docks below the drag handle via `--sheet-sticky-top`.

Preview: https://vega5.tailc47d6.ts.net:8401

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,14 @@
 
 * Releases now deploy the API themselves instead of waiting up to an hour for the server to notice a new image.
 * Your account row stays pinned to the bottom of the mobile dashboard, so you can reach it without scrolling past every recent place.
+* Search suggestions run the full height of the panel instead of stopping half way down with empty space beneath them.
+* The layer picker opens as a sheet you can drag up to full screen again.
 
 ### Fixed
 
 * Panels inside the mobile sheet scroll with the sheet instead of trapping the drag — search results, the timeline, collections, notes and friend and tracker details all pan and expand properly again.
 * The layer picker's tabs stay put while you scroll the list, and the sheet itself drags instead of fighting an inner scrollbar.
+* Lists leave room for the on-screen keyboard, so you can scroll to the end of your results without dismissing it first.
 
 ## [0.10.4] - 2026-09-06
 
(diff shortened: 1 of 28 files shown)


alexwohlbruck:

PR #467, opened 2026-09-07
Transit route detail: camera, stop spine, network dim

alexwohlbruck:
Four fixes to the transit route detail view.

### The camera sometimes stopped short of the whole route

`route-isolation.service` called `mapInstance.fitBounds` directly, with a hardcoded `left: 420` standing in for the panel's width. The LeftSheet slides in over ~300ms, and its `visibleMapArea` watcher calls `map.setPadding()` on every frame of that slide — `setPadding` is a `jumpTo` underneath, so it calls `stop()` and kills the in-flight 800ms fit wherever the ease had got to. Arriving with the sheet already open left nothing to interrupt, which is why it only happened sometimes.

Now it uses `mapService.fitBounds` — the obstruction-aware wrapper the timeline and place views already use. It measures the real visible area instead of guessing, and re-fits once the drawer settles. The `FitBoundsFn` type moved from `timeline-layer.service` to `map.types` now that two services share it.

### The stop list's line missed its dots

The dots were `position: absolute` with no `top`, so they landed at each row's static position (row top + 2px) while the spine and the vehicle markers were placed at row-height/2 — an 8.5px disagreement on short rows, which is the line hanging past the last stop. Everything now derives from one `STOP_DOT_CENTER_Y`, pinned by a test on both row heights.

### The rest of the network is dimmed, not hidden

Isolation in portolan was a filter that narrowed every ribbon layer to the one route. It's a per-feature opacity expression instead: the route keeps full strength where it actually runs at this hour, everything else drops to 0.3, so you can still see where the line crosses the network. Each layer's base colour and opacity are recorded at creation so the dim re-derives from the original rather than stacking on the last dimmed value; Mapbox keeps its alpha folded into the colour, since `line-occlusion-opacity` refuses a data-driven `line-opacity`. The legacy transitland fade moved from 0.15 to the same 0.3.

Stations are the deliberate exception — dots, names and bullet strips are still narrowed to the route rather than dimmed. A label strip at 30% is noise competing with the ones you're trying to read, and labels are the scarce resource on that map.

### Back arrow

Dropped the panel's own back arrow and title row. The sheet's floating button column already shows one whenever there's history, which is what place detail does.

---

`vue-tsc` clean, 2185 tests pass. No screenshots: nothing was running on 5173/5000 locally and I didn't want to start a server — worth an eyeball on the dim level (0.3) and the stations call before merging.


alexwohlbruck:

PR #468, opened 2026-09-07
Dim dark-mode shields

alexwohlbruck:
Highway shields were the brightest thing on the night map and there were far too many of them zoomed out (screenshot below — a wall of white boxes over Queens, most of them motorway exit numbers).

Three changes to the basemap style:

- **Dark-mode shield art.** The sprite sheet now carries a `-dark` cut of every route marker: charcoal plaques with light numerals for state/default routes, and the interstate / US-route markers pulled down to map weight while keeping their legislated hues. The style picks the variant through a new `@shield_variant` token, so light mode is untouched.
- **Exit numbers no longer masquerade as route shields.** The `Highway shield` layer had no `subclass != junction` filter, so exit refs ("22A", "9E", "1W"…) drew as white plaques from z6 while the real junction layer waits until z16. They're now excluded — that alone was most of the clutter.
- **Zoom-gated density.** Zoomed out, only motorways carry shields; trunk joins at z8, primary at z10, and everything (the old behaviour) at z12.

### Light mode (unchanged art, same density gate)

Screenshots live on the throwaway `assets/pr-468` branch; delete it after merge.

All 232 map-style tests pass. Sprites rebuilt and committed (`bun run build:sprite`).


alexwohlbruck:

PR #469, opened 2026-09-07
Fade map bullets by service

alexwohlbruck:
Follows #467, which merged while this was in progress. Two changes, both about the map agreeing with the panel beside it.

## A station's own page fades its map bullets

#467 taught the map to fade a bullet for a line that isn't running, but only fed it from an opened **route**. An opened **station** had the same board data and was only giving it to the header — so the header dimmed the 2 and the 5 at Crown Hts–Utica while the map beside it showed all four lit.

The station page now publishes to the same channel. That channel is keyed by publisher (`route` / `place`) so an opened line and an opened station can each answer without erasing the other on teardown.

Worth stating plainly, because it is the whole reason boards are needed here: the tiles' `acts` masks are a **weekly** timetable indexed by day-of-week. Today is Labor Day — a Monday running a Sunday service — and the masks report Monday's, where the 2 and 5 do serve Utica. GTFS holiday exceptions (`calendar_dates.txt`) aren't in them, so no timestamp can correct it. Only a board knows.

Still limited to stations a board has been read for (an opened line or station). On the plain network map the timetable still answers.

## An isolation's renderer is decided once

A regression I introduced in #467. `render()` became re-entrant so the map could react as boards and alerts settled — but it also held the decision of *which* renderer draws the line, and that test scans the tiles **currently loaded**. So each settle re-asked the question; if the map had panned off the route, the answer came back "portolan doesn't draw this", the reconciler counted three idles, and it tore down portolan's ribbon, stations and bullets for the plain shape-and-circles overlay — mid-view.

Now: who draws is decided once per isolation and never revisited. A later pass may change *what* is drawn (the geometry, when a break is confirmed); it may not change *who*. Reset only on opening another route or closing the panel.

Reproduced and verified: opened the 4, panned to Queens while the boards were landing, waited out the idles, panned back — no `route-detail-*` layers at any point, 72 portolan ribbon layers, 23 portolan stations, Crown Hts–Utica reading `2~ | 3 | 4 | 5~`. Before, it was white dots and plain labels by then.

Tests: 2225 pass, typecheck clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 * On the dark map, the rest of the network stays more visible behind an opened line.
 * An opened line sits on its own track rather than beside it, with stop dots sized to the line and drawn in its colour — no more white bundle markers or dots left in a lane the line no longer runs in.
 * An opened line shows its stop names and connection bullets at every zoom, including the whole-route overview.
-* Route bullets on the map fade for lines that aren't running at a station, matching the stop list. The map used to go by the timetable alone, which on a holiday or a reroute day says every line is running.
+* Route bullets on the map fade for lines that aren't running at a station, matching the stop list and the station header. The map used to go by the timetable alone, which reads the wrong day on a holiday — Labor Day runs a Sunday service, and the map still showed you Monday's.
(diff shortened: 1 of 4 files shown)


alexwohlbruck:

Issue #132, opened 2025-04-16
Realtime lighting


alexwohlbruck:
(Mapbox only) Lighting and map theme should be able to reflect realtime position and color of the sun at the user's location. This should be a user preference setting.

PR #475, opened 2026-09-08
Trains where they are

alexwohlbruck:
Four fixes to the isolated route view. The first two share one root cause.

## Timeline measurement went stale and stayed stale

Rows size to their own content, so the spine and the train markers are placed from measured dot centres. Those centres came out of a Vue `v-for` ref array — which Vue fills but never empties, so a list that *shrank* kept the longer one's tail. Measurement then bailed on the count mismatch and left the old numbers standing, forever.

Two symptoms, both visible on the 4 today, where alerts shorten the list after it first renders:

* every train pinned to the top of the list (`vehicleTopPx` falls back to `spineTop` when the counts disagree)
* the line ran on past the last stop, hundreds of pixels down the page (the whole-length fallback segment drew between two stale centres)

The rows are now read from the DOM (`[data-stop-row]`), which cannot be stale, and measurement runs inside the `flush: 'post'` watcher rather than a tick later — so there is no frame without numbers, and the unmeasured fallback that caused the overshoot is gone.

Measured live: 46 rows, 45 spine segments, the last ending at 2334px — exactly the last dot's centre. Before: one 2749px segment against a 2334px list.

## The 4's map drew 6 trains

barrelman's `findRelatedRoutes` matched on `route_color`, and a colour is a *trunk*: the 4, 5, 6 and 6X are all MTA green. Fixed in barrelman 0.2.24 — related now means the same line's express working.

## Picking a train showed no times

`/transit/trip-stops` returned the feed's raw stop ids, which for the MTA subway are platforms (`250N`); the route's stop list holds stations (`250`), so nothing matched. barrelman 0.2.24 names each stop time's `parentStation`; the store files each time under both.

## Header

The train picker moves up beside the direction picker in the pinned header, reading as the count (`30 trains active`) until one is picked. The timeline gets its own stacking context, so a marker no longer scrolls over the header.

---

Needs barrelman ≥ 0.2.24 (alexwohlbruck/barrelman#132) for the last two.

Full web suite green (2234 tests). Two new timeline tests: the spine must not outrun the last stop when the list shortens, and a train must be placed by its distance along the route.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@
 
 ### Fixed
 
+* Live trains sit where they are on a route's timeline. They were pinned to
+  the top of the stop list, and the line ran on past the last stop, whenever
+  the list shortened under the panel
+* A route's map shows only that line's own trains. The 4 drew the 5s and 6s
+  too, since all three are the same green
+* Picking a train fills the stop list with its times again
+* The train picker sits with the direction picker in the pinned header, and a
+  train no longer rides over that header while scrolling
+
 ## [0.11.1] - 2026-09-07
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


alexwohlbruck:

PR #478, opened 2026-09-08
Stops as places

alexwohlbruck:
A directions stop is a place, and now looks like one everywhere it appears — with the metadata to back that up.

### One derivation, three surfaces
`waypointToDisplay` (in `lib/place-display.ts`) is the single answer to "what mark does this stop get": its own POI glyph, the locate mark for current location, or the pin for a plain point. The waypoint fields, the trip timeline and the map markers all read it. They used to each decide separately and drifted — a stop drawn as a restaurant on the map was a numbered disc in the input, and an uncoloured fallback pin came out cobalt in the timeline against a neutral plate on the map.

Numbered discs and the destination flag are gone. The origin keeps an open ring, since it's the one stop that isn't a place.

### Trip timeline
The stop's glyph **is** the rail node — drawing a dot on the rail *and* an icon beside it was two marks for one thing. Stops are plain rows (name, type, arrival time), no boxes: a muted card on a muted panel had no edge to see, and boxes fought the segments between them. Still `PlaceCard` (`plain`, no icon), so icon/type/colour/link derivation stays shared.

The hero is pinned via `SheetHeader`, same contract as the transit route detail.

### Map
Stops wear the place's own marker — same plate, glyph and ring as a search result — and carry the stop's name, lettered exactly as the map's own POI labels are: the icon's glyph colour on the basemap's `poi_halo`, Geist Medium at 13px, 1.1em below the icon's centre. `MARKER_HALO` moved out of `marker-paint.ts` so the DOM label and the style layer read one value.

### Metadata through the flow
- `autocompleteResultToPlace` no longer drops the row's resolved icon, icon pack, category or transit refs — and `WaypointInput`'s duplicate copy of it is gone.
- Waypoints hydrate in the background off a store watcher, so every path gets it — the picker, a map click, a shared link, the place page's Directions button.
- Shared links carry a `wpid` per waypoint, so a reload restores what each stop *is*, not just its label.
- `RouteWaypoint.place` is a declared field rather than an `as any` the client wrote in anyway.

### Touch targets
"Details · N steps", "N stops · 12m" and "Advanced" are full-width controls on the card surface — 36px on a mouse, 44px under `pointer-coarse`.

### Also
- `PlaceCard`/`ItemRow` gain a `plain` variant (no surface, for lists that draw their own structure) and a `title-trailing` slot.
- The waypoint picker's dropdown rows are `PlaceCard`s, so a maki glyph stays a maki glyph.
- Current location shows as a chip in the field rather than editable-looking text.
- `ElevationChart` gets `isolate` — its internal `z-20` badges were tying with the sticky header and painting over it.

Tests: added coverage for the `wpid` round-trip, `shareableWaypointId` and `autocompleteToDisplay`. `vue-tsc` clean, 2243 pass.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Picking your current location as a stop shows it as a chip in the field,
   so it reads as a value rather than typed-in text
 * "Details" and "Advanced" toggles are big enough to tap on a phone
+* A trip's duration and times stay pinned while its timeline scrolls
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@
 
 ### Changed
 
-* Stops in a trip's timeline show as place cards — the place's own icon,
-  category and hours, with the scheduled time beside the name
-* Route stops on the map wear their own POI icon instead of a numbered dot
+* A trip's timeline draws each stop as the place it is — the POI's own icon on
+  the route line, its name and type beside it, and the scheduled time
+* Route stops on the map wear their own POI icon and name instead of a
+  numbered dot
 * Picking your current location as a stop shows it as a chip in the field,
   so it reads as a value rather than typed-in text
 * "Details" and "Advanced" toggles are big enough to tap on a phone
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,17 @@
 
 ### Changed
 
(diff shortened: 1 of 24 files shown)


alexwohlbruck:

alexwohlbruck:

PR #479, opened 2026-09-08
Transit off clears the map

alexwohlbruck:
Switching the Transit layer group off left the whole network painted — every
ribbon, station and label stayed on the map while the switch, the class
toggles and the service-time control all read "off".

## What was wrong

Two bugs, both from reading `isStyleLoaded()` as "the style is usable". It is
not: it is a claim about *every* source cache in the style, and portolan
installs one vector source per pyramid (90 of them), so it reads false
whenever a single tile is in flight — which is exactly the moment someone
touches the switch. The file already documents this trap for hydration
(`hydrationReady`); these two call sites had not learned it.

1. **Teardown skipped its removal.** `teardownPortolanTransit` only called
   `removeAll()` `if (map.isStyleLoaded())`, then set `map = null` — so when
   the guard was false the layers stayed and nothing was left that could come
   back for them. Instrumented in the running app: at the moment of the
   toggle the guard read `false`, `removeLayer` was called 0 times, and all
   99 portolan layers remained.

2. **The rebuild waited for an `idle` that never came.** `sync()` deferred its
   build to `map.once('idle')` when the style was "not loaded". A map that has
   already settled never fires another idle, so switching the group back on
   drew nothing until the rider happened to pan — and the guard flag that
   deferral set was cleared only by that same handler, so one missed idle
   latched the feature off for the rest of the session. This was invisible
   before, because the layers were never removed in the first place.

## The fix

* Teardown removes unconditionally (`removeAll` keeps its own no-style guard).
* Building gates on `styleReady()` — the style *parsed*, which is all
  `addLayer` needs and all that the label/building anchors need — instead of
  fully fetched, and retries on a short bounded timer rather than an event
  that may never arrive.

## Verified

In the running app, driving the group toggle and counting portolan layers and
sources on the style:

| | before fix | after fix |
|---|---|---|
| toggle on | 99 layers / 9 sources | 111 layers / 10 sources |
| toggle off | **99 layers / 9 sources** | **0 / 0** |
| toggle on again | (never removed) | full network back, no pan needed |

Three consecutive cycles on `dev` and two on this preview, clean each time.

New unit tests cover both: they fail against the old service (removal skipped,
`once('idle')` parked) and pass against the new one. Full suite: 2238 passing,
`vue-tsc` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Fixed
 
+* Switching the Transit layer off clears its lines and stations from the map,
+  and switching it back on draws them again without waiting for a pan
+
 ## [0.11.2] - 2026-09-07
 
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,12 @@
 - If a Linear ticket was linked for the relevant work, update the status of the ticket as work progresses.
 - Use `bun` over `npm` for package management.
 - Commit messages: short (5-20 words), distinct logical commits.
-- Keep code structure clean, modular, and dry. Use concise, straightforward naming conventions and move code to appropriate modules when it isn't in the correct place. Add comments ONLY when the code is not intuitive at a glance or to convey important context/information. Do not write paragraphs of comments explaning each iterative change.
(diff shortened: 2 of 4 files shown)


alexwohlbruck:

PR #480, opened 2026-09-08
Offline foundation

alexwohlbruck:
Foundation for intentional offline behavior: offline is a state, not an error. Implements the offline-improvement brief end to end (except downloadable map regions, explicitly out of scope).

**Preview:** https://vega5.tailc47d6.ts.net:8404 · **Docs:** alexwohlbruck/parchment-docs#10

## What's here

- **PWA** (`vite-plugin-pwa`, injectManifest): manifest + generated icons (`bun run build:pwa-icons`), custom `src/sw.ts` precaching only the app shell (~9 MB — full dist is 24 MB), runtime LRU caches for lazy chunks, glyphs/sprites/models, and **map tiles** (`/proxy/*` + satellite/terrain, 2500 entries/30 days, auth params stripped from keys). Tile cache-buster is now the app version instead of `Date.now()` so caches survive across sessions. Registered in production web only, never Tauri.
- **Connectivity state** (`lib/connectivity.ts` + `useConnectivity()`): browser events + request-outcome evidence + bounded server probe. `online / offline / reconnecting`. Realtime reconnects instantly on it instead of waiting out backoff.
- **Error classification** (`lib/network-errors.ts`): offline / unreachable / timeout / 4xx / 5xx / cancelled / unknown, stamped on every axios error. Offline + cancelled are silent; unreachable/timeout collapse into one toast; timeouts get their own message instead of the generic "Network error".
- **Offline reads suppressed** at the request interceptor (typed `OfflineRequestError`, `allowOffline` opt-out) — no 15s timeout burn per feature, existing catch-with-fallback services degrade quietly.
- **Offline boot fixed**: background session validation no longer signs out + wipes caches on a network failure; bootstrap survives failed fetches.
- **Mutation queue** (`stores/sync.store.ts` + `lib/sync/`): persistent, ordered replay, temp-id remapping, retriable-vs-definitive failure semantics, rollback + failed state + retry/dismiss. Bookmarks (save/update/remove) integrated as the reference: optimistic local apply, offline temp ids, store refreshes preserve unsynced rows. Queue clears on sign-out.
- **Sync UI** (Linear-style): quiet sidebar pill ("Offline · n"), hover panel listing pending/syncing/failed with per-item cancel/retry/dismiss. Desktop banner slot + mobile sheet footer.
- **Offline empty states**: shared `EmptyState` gains an `offline` mode (swapped copy, hidden network actions, retry that probes connectivity); wired into Collections, Routes, Layers, Canvases, and search results, each refetching via `onReconnected`.

## Verified

- `vue-tsc` clean, full vitest suite green (2259 tests; 16 new for classification/connectivity, 9 for the queue).
- Production build emits `sw.js`, manifest, 19-entry precache.
- Live on the preview: with the device offline and queued changes, the sidebar shows the "Offline · 2" pill; its panel lists "Pending changes / Your changes will sync when you reconnect" with cancel on pending rows and retry/dismiss + error on the failed row.

## Notes for review

- The service worker only runs in production builds — branch previews (vite dev) exercise everything except SW caching.
- `updateBookmarkCollections` in the collections store conflates the two sides of the bookmark↔collection relation (pre-existing); the offline path mirrors the online path's behavior rather than changing it here.


alexwohlbruck:

alexwohlbruck:
### Fixed: offline was still shouting

Two bugs reported on the preview — a technical `Request suppressed while offline: get /places/details` toast over a blank place panel, and `Failed to fetch collections` when opening the library.

Root cause was mine: suppressing the error at the axios interceptor only silenced the *global* toast. Nearly every service catches its own fetch failure and raises a message of its own, so ~80 `toast.error` call sites across 26 files fired regardless of connectivity.

Fixed systemically rather than per-file:

- **`web/src/lib/toast.ts`** — all toasts now route through one wrapper that drops `toast.error` while offline. Success/info/warning pass through (an optimistic "Saved" offline is true and worth showing). Escape hatches: `{ force: true }` to show anyway, `{ cause: error }` to drop on the error's own merits (covers the race where connectivity returns before the catch block runs). All 15 `vue-sonner` imports + 1 dynamic import rewritten; only the `Toaster` component still imports sonner directly.
- **`OfflineRequestError.message` is now plain English** ("You're offline") — some catch blocks render `error.message` straight into the UI, so the internal description should never have been user-visible. `method`/`url` carry the detail for the console.
- **The place panel rendered nothing when there was no place** — that was the blank area in the screenshot. It now shows the offline empty state (or "Place unavailable" when online), with retry wired to refetch.
- **Friendlier copy**: "You're offline" / "Go back online to view this" / "Try again".

Verified on the preview with the device offline: the place route shows the empty state with no toasts, and opening Collections renders cached collections with no error. 7 new tests lock the policy in (including "shows a genuine server error that arrived after reconnecting", so this doesn't silence real failures). Full suite green: 2266 tests.

PR #482, opened 2026-09-08
Controls ride the sheet

alexwohlbruck:
## What

On mobile, the bottom-right map controls (street view, layers, locate) were covered whenever a bottom sheet was open — searching the map meant losing the geolocate button entirely.

## The change

The control cluster in `Map.vue` now reads the sheet's live bounds (the same `map-content-sheet` entry the street-imagery peek uses) and translates up to sit 8px above the sheet's top edge. The bounds publish every frame during drags, so the buttons track the sheet in real time. Once the sheet covers more than 70% of the screen the map is effectively hidden, so the cluster fades out and stops accepting taps.

Desktop is untouched; with no sheet open the buttons sit exactly where they did before.

## Verified

On the slot-1 preview at iPhone 12 Pro size (390×844):

- Search results sheet at half height (top y=422): buttons lift to y=266–414, bottom edge 8px above the sheet, and an `elementFromPoint` hit test on the locate button resolves to the button.
- Sheet dragged to full height: the cluster's opacity settles at 0 with `pointer-events: none`.
- Sheet closed: buttons back at their resting spot (y=624–772) with no inline styles left over.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@
 
 ### Fixed
 
+* On mobile, the layer and locate buttons stay reachable when a sheet is open —
+  they ride its top edge instead of hiding behind it
+
 * A shared directions link keeps what each stop actually is. Reloading one
   used to leave every stop as a bare name
 * Escape closes a trip's details on the first press. It used to take two — the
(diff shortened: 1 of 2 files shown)


alexwohlbruck:

PR #487, opened 2026-09-09
Fix sheet snap overshoot

alexwohlbruck:
Dragging the bottom sheet up from peek and releasing near middle always expanded to full. In the vaul-vue fork's `onRelease`, the guard for "already on the last snap point" tested the computed ref object (`isLastSnapPoint`) instead of `.value`, so it was always truthy — any upward release above the velocity threshold snapped to the last point, making the one-step branch unreachable.

Fixed in the fork (alexwohlbruck/vaul-vue#1) and published as `@alexwohlbruck/vaul-vue@0.4.5`; this PR bumps to it.

Test on Chrome mobile: open a sheet with 3 positions, drag peek → middle with a natural flick, release just before middle. It should land on middle, not full.


alexwohlbruck:

PR #488, opened 2026-09-09
Reorganize web app by module

alexwohlbruck:
Restructures `web/src` around modules, deletes what nothing references, and fixes six defects the audit turned up along the way. Every commit is independently green: `vue-tsc` clean and 138 test files / 2304 tests passing at each step.

## Why

An audit of all 619 source files (plus the 295-file design system) found the app organised by **layer** rather than by module, so every feature was smeared across five directories. `src/lib` held 153 files at one level, `views/` held ~20 components no route renders, and `components/map/layers/` held eight plain-TS runtime classes with no template.

## Bugs fixed

- **Map tools broke after any style change.** `unbindMapEvents()` called mitt's `off(type)` with no handler, which removes *every* listener for that type — so an engine switch or locale change silently tore down the `style.load` listeners owned by MeasureTool, RadiusTool, IsochroneTool and the shade tuner. Those three tools also never removed their own listeners on unmount.
- **Unit system never synced from the server.** A cast to `{ value }` assigned `.value` on a Pinia-unwrapped string — a no-op where swallowed, a `TypeError` in `updatePreferences`.
- **Collections created from the library lost their icon pack.** The form dialog was invoked from three places, each mapping params by hand; the library list had drifted and omitted `iconPack`.
- **`/settings/developer` was reachable in production**, with working impersonation and location-broadcast controls. Only the nav entry was DEV-gated, not the route.
- **`ColumnMeta` augmented `@tanstack/react-table`** while the app uses `@tanstack/vue-table`, so it was inert. Fixing it surfaced a latent type error in `RoleDetail.vue`.

## Structure

| | Before | After |
|---|---|---|
| `src/lib` flat files | 153 | 29 |
| Components under `views/` | ~20 | 0 |
| Non-component `.ts` in `components/` | 8 | 0 |
| Dead files | — | 69 removed (~3,900 lines) |

- `lib/` grouped into `canvas/ directions/ geo/ identity/ layer-editor/ map/ measure/ place/ realtime/ search/ street-view/ transit/ weather/`. `map-style/` was two packages under one name — the basemap pipeline plus ~1,450 lines of layer-editor domain model used only by library UI.
- Six E2EE components moved out of `friends/` into `identity/`: none imported `friends.store`, all six import `identity.store`.
- The sheet system (spread across five directories, sharing four provide/inject contracts) is now one module.
- Map layer runtime moved to `services/layers/`; the layer editor UI moved to `library/layers/`; the layer *selector* moved from `navigation/` into `map/layers/`. `street-view/` and `weather/` became modules.
- `place/` lost its four one-file folders and its transit UI; the two largest transit components lived there while `transit/` held only leaf primitives.

## Deduplication

haversine existed six times in three formulations; `capitalize` nine times; `formatCountdown` three times, all reimplementing something `lib/transit-departures` already exported.

## Naming

`eventBus.ts`/`vaulChromeWorkaround.ts` were the only camelCase files among 150. `storage.ts` held one serializer, `route-prefetch.ts` prefetches *router* chunks. `Map.vue` and `Layers.vue` each existed twice and were imported under their own filename. List components are now singular (`RouteList`, not `RoutesList`).

## Not done here

Deliberately left, with rationale:

- **Route imports stay static** — lazy chunks would regress offline behaviour.
- **i18n still bundles both locales** (~196 KB, half wasted per user). Lazy locale loading is the single biggest remaining bundle win but carries the same offline question, so it wants your call.
- **The 8.1 MB entry chunk** still has no `manualChunks` strategy; both mapbox-gl and maplibre-gl ship to every user.
- **Large-file splits not attempted**: `portolan-transit.service.ts` (93 KB, 13 concerns), `TripDetail.vue` (83 KB, 10 concerns), `maplibre/mapbox.strategy.ts` (72% duplicated), `map.service.ts` (10 concerns), `layers.store.ts`. Each deserves its own PR.
- **`lib/key-pinning.ts` and `lib/metadata-crypto.ts`** are E2EE features with tests but no callers — kept, since deleting unwired security code should be your decision.

`CLAUDE.md` gains a Frontend architecture section encoding the conventions whose absence allowed this drift.


alexwohlbruck:

alexwohlbruck:
---

## Update: large-file splits

Second pass, 8 more commits. Same discipline — typecheck clean and the full suite green at every step. These files had **zero test coverage**, so I did behaviour-preserving extractions and wrote tests for the logic as it came out: **2,368 tests now, up from 2,304**.

### Two more bugs, both found by the new tests

- **Sheet peek could overshoot the next detent on a device with a home indicator.** The clamp that exists to keep snap points ascending (vaul's drag maths assumes it) ran *before* the safe-area inset was added to the peek. A measured peek of 900px clamped to 376px, then gained 30px and landed above the 400px detent. Clamping last fixes it.
- **`bareRouteId` was implemented twice** with different regexes, and the network-dim values were tuned as two separate constant pairs in two files that call each other.

Also fixed: `proxyBase` was written three times across the portolan files and `ensureRegions` twice **with separate caches**, so opening the map and opening a station header each fetched the pyramid index.

### What came out

| File | Before | After | Extracted to |
|---|---|---|---|
| `command.store.ts` | 746 | 436 | `services/search-suggestions.service.ts` |
| `route-detail.store.ts` | 838 | 646 | `lib/transit/route-projection.ts` (+14 tests) |
| `layers.store.ts` | 1305 | 1244 | `lib/map/layer-tree.ts` (+9 tests) |
| `TripDetail.vue` | 1917 | 1811 | `lib/directions/trip-display.ts` (+19 tests) |
| `BottomSheet.vue` | 1018 | 985 | `components/sheet/snap-points.ts` (+17 tests) |
| `map.service.ts` | 1695 | 1622 | `lib/map/map-padding.ts` (+5 tests) |
| strategies | 3219 | 3046 | 16 identical methods hoisted to `MapStrategy` |

### On the strategy duplication

The audit claimed 72% duplication / ~1,400 deletable lines. That was a **line-level** match counting `}` and `const map = this.mapInstance`. Measured per method: **17 of 39 shared methods are byte-identical**, 22 genuinely differ (the label toggles use `setConfigProperty` on mapbox and layer-group visibility on maplibre — not unifiable). 16 are now hoisted; `getBounds` is left duplicated because its multi-line return type makes the move non-mechanical.

### Still not split, and why

`portolan-transit.service.ts` is down only 10 lines and remains 2,453. It holds **~25 module-level mutable `let`s** shared across all 13 concerns, with bidirectional calls between the isolation, hydration and filter paths — extracting any slice creates a circular import or silently splits state. With no test coverage on a live map renderer, a blind restructure is how you ship a subtly broken transit map. It needs tests characterising the mount/hydrate/isolate cycle first; that is its own PR.

Same reasoning, smaller scale, for the rest of `map.service.ts` (a `createSharedComposable` closure whose parts share `mapStrategy`, `paddingHold` and nine watchers) and `CanvasEditor.vue`.

alexwohlbruck:
# Reorganization map

## 1. The shape of the problem

The app was organised by **layer**, so one feature lived in five places at once.
Transit, before — 44 files, scattered:

```
components/                          lib/                     stores/
├── transit/          ← 9 leaf bits  ├── transit.ts           ├── portolan.store
├── place/details/    ← PlaceTransit ├── transit.utils.ts     ├── transit-alerts
├── place/pages/      ← 16KB board   ├── transit-alerts.ts    └── transit-vehicles
├── directions/timeline/ ← Departure ├── transit-bullets.ts
├── navigation/       ← ServiceTime  ├── transit-departures
├── map/              ← StopMapIcon  └── transit-reachability
└── map/layers/       ← 36KB vehicles
```

`transit/` held only the small leaf primitives. The 16 KB stop board and the
30 KB route panel lived under `place/`, and the 36 KB vehicle layer — plain
TypeScript, no template — lived under `components/`.

## 2. After

```
components/transit/                  lib/transit/             services/layers/
├── alerts/        6 files           ├── transit.ts           ├── features/
├── bullets/       1                 ├── transit-alerts.ts    │   ├── transit-vehicles-layer
├── departures/    5                 ├── transit-bullets.ts   │   └── portolan/  (11 files)
├── routes/        1                 ├── transit-departures   └── markers/
└── TransitServiceTimeControl.vue    ├── transit-reachability
                                     ├── route-projection.ts  ← new, tested
                                     └── …
```

Same 44 files. You can now find any of them from the feature name.

## 3. Layer boundaries

```
      views/          routed screens only
        │             (22 non-view components moved out)
        ▼
    components/       templates only
        │             (8 plain-.ts layer classes moved out)
        ▼
   composables/       caller-scoped reactive state
        │
        ▼
      stores/         named singleton state
        │
        ▼
    services/         I/O + orchestration
        │
        ▼
       lib/           pure, framework-free
```

Arrows point one way. A `lib/` file importing a store, or a service importing
a `.vue`, is now a defect the CLAUDE.md rules name explicitly.

## 4. Directory counts

| | before | after | |
|---|---|---|---|
| `lib/` top-level files | **96** | **22** | grouped into 21 packages |
| `lib/` packages | 10 | 21 | canvas, directions, geo, identity, layer-editor, map, measure, place, realtime, search, street-view, transit, weather … |
| `views/` files | 65 | **43** | 22 were never routed |
| `components/` files | 552 | **513** | dead + design-system removals |
| single-file dirs | `config/` `filters/` | — | retired |

New component modules: `identity/` `shell/` `timeline/` `weather/`
Retired: `example/` `layouts/`

## 5. `src/lib` before → after

```
BEFORE                          AFTER
lib/                            lib/
├── 96 loose files              ├── 22 loose files  (api, toast, utils,
│   transit-* ×8                │                    connectivity, time …)
│   canvas-*  ×5                ├── canvas/         ├── measure/
│   place-*   ×4                ├── directions/     ├── place/
│   geo-*     ×3                ├── geo/            ├── realtime/
│   passkey-* ×3                ├── identity/       ├── search/
│   *-crypto  ×5                ├── layer-editor/   ├── street-view/
│   …                           ├── map/            ├── transit/
└── 10 tidy subpackages         │                   └── weather/
    (the pattern already        └── + the original 10
     existed, unapplied)
```

## 6. Where the split logic landed

| Extracted from | To | Tests |
|---|---|---|
| `command.store` 746→436 | `services/search-suggestions.service.ts` | — |
| `route-detail.store` 838→646 | `lib/transit/route-projection.ts` | +14 |
| `TripDetail.vue` 1917→1811 | `lib/directions/trip-display.ts` | +19 |
| `layers.store` 1305→1244 | `lib/map/layer-tree.ts` | +9 |
| `BottomSheet.vue` 1018→985 | `components/sheet/snap-points.ts` | +17 |
| `map.service` 1695→1622 | `lib/map/map-padding.ts` | +5 |
| both map strategies | 16 methods → `MapStrategy` | — |

Still oversized, deliberately: `portolan-transit.service.ts` (2453), whose 25
shared mutable module globals need characterisation tests before any split.

PR #489, opened 2026-09-09
Tidy web architecture

alexwohlbruck:
Second cleanup pass over `web/src`, following #488.

Pass 1 reorganised `lib/`, `views/` and `components/`, then wrote the resulting rules into `CLAUDE.md`. This pass audits the codebase **against those rules** and fixes what fails — including two rules that turned out to be wrong.

225 files, +594 / −903. Every commit typechecks and runs the full suite.

## The last flat dumping ground

Pass 1 fixed `lib/` (96 → 22 root files) and never touched `composables/`, which had grown to the same shape.

```
BEFORE                       AFTER
composables/                 composables/
├── 71 loose files           ├── 16 loose files  (cross-cutting only:
│   useCanvas*      ×8       │   useAbortController, useClipboard,
│   useFriend*      ×5       │   useHotkeys, useUnits, useUpdater…)
│   useSheet/Drawer ×5       │
│   useTransit*     ×4       ├── canvas/    8   ├── sheet/     5
│   useMap*         ×4       ├── map/       8   ├── place/     3
│   …                        ├── friends/   5   ├── settings/  2
└── 0 subpackages            ├── transit/   4   ├── integrations/ 2
                             └── trackers/  1
```

## Rule violations found and fixed

**`lib/` must not reach into stores, services or components** — was 6, now 0.

| What | Why it was wrong | Where it went |
|---|---|---|
| `map-providers/*.strategy.ts` (3 files, ~3.1k lines) | Plain `.ts` classes with no template, living in `components/`, imported by 20 files that are nearly all services | `services/map/providers/` |
| `lib/map/layer-group.ts` (820 lines) | In `lib/` but imports a strategy and a `.vue` marker | next to its only three consumers |
| `lib/place/place-colors.ts` | A pure-looking wrapper whose only job was calling `useCategoryPaletteStore()` | `services/place/` |
| `lib/place/place-display.ts` | Transitively store-bound through the above | `services/place/` |
| `ChipOption` in `lib/search/` | A `ui/` primitive's type inverting the layering | `FilterOption` in `types/search.types.ts` |
| `DefaultUserStateRow`, `DefaultStateType` | Service-owned types imported by `lib/` | `types/map.types.ts` |

**No `.utils.ts`** — was 15, now 0. Renamed for what they actually hold, not blanket-stripped: `time.utils` → `time-format`, `timezone.utils` → `timezone-warning`, `search.utils` → `search-result`, `place-open.utils` → `place-hours`.

Three were misnamed *because* they held unrelated things, so the rename was a split:

* `transit.utils` → `route-colors` + `transitland-layers` + types
* `place.utils` → `place-route` + `place-details`
* `map.utils` → `mapbox-style-compat` + `osm-feature-id`

**No two files sharing a name** — `library/EmptyState.vue` → `LibraryEmptyState.vue`, `admin/Users.vue` → `UserTable.vue`. Both were children shadowing the name of the thing that rendered them.

## Dead code

Splitting `map.utils.ts` and `transit.utils.ts` exposed two dead blocks that their filenames had been hiding:

* **`parseOpeningHours`** (~155 lines) — zero callers. The server owns hours parsing; the web reads `resolveOpeningStatus` instead. It also carried a private `OpeningStatus` type duplicating the live one in `place-hours.ts`.
* **The entire transit colour module** — `TRANSIT_ROUTE_TYPE_COLORS`, `getRouteTypeColor`, `isValidRouteColor`, `normalizeRouteColor`, the `TransitRouteType` enum and the `TransitRoute`/`TransitStop` tile interfaces. All unreferenced.

The second one is worth naming, because it's a repeatable failure mode: **a live `getRouteColor` in `lib/transit/transit.ts` shares its name with the dead one that sat beside it.** Every grep for the symbol hit the live one, so the dead file read as used for as long as it existed. There were also two unrelated `TransitRouteType` definitions — an enum here and a string union in `types/multimodal.types.ts`. That's now a rule.

Also hoisted the `getBounds` override that pass 1 left duplicated byte-for-byte in both strategies.

## Two rules I got wrong

Pass 1 wrote these; enforcing them here showed they don't hold, so they're amended rather than worked around:

* *"`components/<module>/` never holds plain `.ts`"* — too absolute. Component-scoped `context.ts` / local `types.ts` are ordinary Vue practice. The rule now targets what actually went wrong: a class or service the rest of the app calls.
* *"`components/ui/` never imports a store or service"* — unworkable as written. `Kbd.vue` resolves hotkeys and is consumed by `ui/sidebar` itself, so moving it only relocates the violation; `ItemIcon.vue` needs `isDark` to tint. The rule now draws the line where it belongs: `ui/` may read **ambient app context** (theme, hotkeys), never a **domain module** (place, transit, directions, library).

No `CHANGELOG.md` entry — this pass is entirely internal, which the changelog rules exclude.

## Verification

`vue-tsc --noEmit` clean · **141 files / 2368 tests pass** · production build clean · app boots on the preview with the map rendering and no new console errors.

### One pre-existing bug found, not fixed here

The basemap logs `Unknown expression "!in"` from `setBasemapTransitPoisVisible`, because `lib/map-style/spec.json` still uses legacy Mapbox filter syntax that MapLibre's `setFilter` rejects — so toggling basemap transit POIs silently fails on MapLibre. Confirmed pre-existing: this branch touches neither `spec.json` nor that method (the only change to `portolan.store.ts` is an import path). Left for its own PR.

### Deliberately not done

`portolan-transit.service.ts` (2,453 lines) still needs characterisation tests before it can be split safely — unchanged from #488's reasoning. Same for the rest of `map.service.ts` and `CanvasEditor.vue`. `stores/` stays a flat list of 26 `*.store.ts`: the filename is the module, so it reads fine as-is.


alexwohlbruck:

PR #490, opened 2026-09-09
Flag degraded integrations

alexwohlbruck:
## Problem

When a user's integration endpoint (e.g. a self-hosted Dawarich) is down, every location-history request pops an "External service request failed" error toast. The failure is the integration's, not the user's current task — the toast is noise and gives no path to a fix.

## Change

A generic degraded-integration mechanism, opt-in per request:

- Any API request tagged with `integrationId` in its axios config now routes upstream failures (502/503) into a persisted degraded flag (`web/src/lib/integration-health.ts`) instead of the global error toast. A successful tagged request clears the flag automatically.
- The Dawarich location-history calls are the first tagged call sites; future integrations opt in with the same one field.
- Settings → Integrations shows a warning badge and an amber "Not responding" status dot on the degraded integration's tile, and the settings nav marks the Integrations page with a dot.
- The timeline's inline error links to Settings → Integrations when the integration is degraded.
- Saving the integration's config or a successful connection test clears the flag.

The flag is per-device (localStorage): health of a user's own endpoint isn't server state, and Dawarich credentials are e2ee anyway.


alexwohlbruck:

PR #491, opened 2026-09-09
Reset clears e2ee integrations

alexwohlbruck:
## Problem

`resetUserIdentity` wipes every `encrypted_user_blobs` row, but never touches the `integrations` table.

A `scheme='user-e2ee'` integration (currently Dawarich) keeps its credentials in a personal blob under `integration-config:<id>`; the `integrations` row itself holds only metadata, with `config_ciphertext`/`config_nonce` NULL. So after a reset the config blob is gone while the metadata row survives:

- `GET /integrations/configured` still returns the row. No blob matches, so `encryptedConfig` stays undefined and the client hydrator falls through to `config: {}` — the tile renders as connected but unconfigured.
- **Reconnecting fails.** The unique index `integrations_user_scheme_uq` on `(user_id, integration_id, scheme)` still matches the stale row, so a fresh connect hits the 23505 path and throws `IntegrationSchemeConflictError`. The user is stuck with no in-app hint that they have to disconnect first.

`deleteIntegration` already pairs the row delete with the blob delete in one transaction, precisely to avoid orphans. Reset just never did the mirror of that.

Same root cause, second table: `routes` is top-level (no cascade reaches it from `collections`) and its `metadata_encrypted` is under the user key for *both* schemes, so every saved route survived a reset as a permanently unreadable row.

## Change

Inside the existing reset transaction:

- Delete `integrations` where `user_id = ?` **and** `scheme = 'user-e2ee'`. `server-key` rows are deliberately left alone — their config is encrypted under `PARCHMENT_INTEGRATION_ENCRYPTION_KEY`, not the user's seed, so they keep working and there is nothing to recover. Wiping them would needlessly disconnect e.g. a linked OSM account.
- Delete `routes` where `user_id = ?`, matching how collections are already handled.

No adapter-cache eviction is needed: `createIntegration` skips `initializeIntegration` for `user-e2ee`, and startup skips them too, so those rows are never in `integrationManager`. Keeping the reset pure-DB keeps it inside the single transaction.

## Tests

`identity-reset.service.test.ts` gains `routes` and `integrations` to its table list, plus a test asserting the integrations delete is scheme-scoped to `user-e2ee`. The drizzle mock now stubs `and` and captures the `scheme` predicate.

Both tests were confirmed to fail against the pre-fix service. `bun run test:unit` is green (3982 passed, 0 failed).

## Not covered

`planned_trips`, `layers` and `user_vehicles` have no seed-encrypted columns, so they are genuinely unaffected. No UI change — this is server-side cleanup, so there is nothing to screenshot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Collections created from the library keep the icon you picked
 * The developer page is no longer reachable in released builds
 * Resetting your encryption key now clears the integrations that relied on it, so they can be connected again afterwards
+* Integrations stranded by an earlier encryption key reset are cleared, so they can be connected again
 * Resetting your encryption key now removes saved routes, which were left behind unreadable
 
 ## [0.11.4] - 2026-09-08
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 * Your unit system preference now applies when signing in on a new device
 * Collections created from the library keep the icon you picked
 * The developer page is no longer reachable in released builds
+* Resetting your encryption key now clears the integrations that relied on it, so they can be connected again afterwards
+* Resetting your encryption key now removes saved routes, which were left behind unreadable
 
 ## [0.11.4] - 2026-09-08
 
--- a/server/drizzle/0057_orphaned_e2ee_integrations.sql
+++ b/server/drizzle/0057_orphaned_e2ee_integrations.sql
@@ -1,0 +1,17 @@
+-- Accounts reset before the reset path cleared them are left with a
+-- user-e2ee integrations row whose config blob is gone. The row reads as
+-- configured, decrypts to nothing, and blocks a reconnect via
+-- integrations_user_scheme_uq.
+--
+-- The age guard exists because a create is two steps: the metadata row
+-- lands first, the client PUTs the blob after. A row minutes old may be
+-- mid-create rather than orphaned.
+DELETE FROM "integrations" i
+WHERE i."scheme" = 'user-e2ee'
+  AND i."user_id" IS NOT NULL
+  AND i."created_at" < now() - interval '1 hour'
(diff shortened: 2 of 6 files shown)


alexwohlbruck:

alexwohlbruck:
## Reproduced live, plus a migration for accounts already broken

Confirmed against the dev database. User 1 has a `dawarich` / `user-e2ee` row whose `integration-config:dawarich` blob is gone — the tile reads as configured and the edit form opens with empty Base URL and API token:

| database | slot | orphaned row? |
|---|---|---|
| `parchment` | 0 (dev) | yes |
| `parchment_p3` | 3 | no — blob still present (cloned pre-reset) |
| `parchment_p5` | 5 | yes (cloned post-reset) |

The reset path is reached from `RecoveryKeyDialog.vue` → "Lost your recovery key?" → `resetIdentity()` → `POST /users/me/identity/reset`, which is what this PR fixes.

Key rotation is a separate path and is **not** affected: `web/src/lib/identity/km-rotation.ts` enumerates `/me/blobs` generically, so `integration-config:*` blobs are re-encrypted along with everything else.

### Added: `0057_orphaned_e2ee_integrations.sql`

The service fix only prevents new orphans; accounts already reset keep theirs, and the unique index still blocks a reconnect. This migration deletes `user-e2ee` rows with no matching config blob.

The age guard (`created_at < now() - interval '1 hour'`) is load-bearing: a create is two steps — the metadata row lands first, the client PUTs the blob after — so a row minutes old may be mid-create rather than orphaned.

Verified transactionally against real data:

- `parchment_p5` (orphaned): `DELETE 1`, e2ee rows 1 → 0
- `parchment_p3` (blob present): `DELETE 0`, healthy row untouched
- `server-key` rows: 18 before and after, both databases
- A freshly-inserted `user-e2ee` row with no blob survives the guard

PR #492, opened 2026-09-09
Link setup warning to settings

alexwohlbruck:
## Problem

Clicking a `user-e2ee` integration tile (Dawarich) before identity setup is done shows a warning toast that names the destination in prose — "Set up your encryption keys in Settings → Encryption keys before configuring this integration." — but gives no way to get there. The user has to read the path, dismiss the toast, and navigate by hand.

This is also exactly what a user sees right after an identity reset, when their keys are gone and every e2ee integration needs reconnecting.

## Change

- The toast gets a **Set up** action that routes to `/settings/account#identity`, following the existing pattern in `Passkeys.vue` (`refreshKeysTitle` → `requestRotateKeys`).
- The message drops the now-redundant path: "Set up your encryption keys before configuring this integration." Both `en-US` and `es-ES` updated, plus a new `setupRequiredAction` label.
- `IdentitySettings.vue`: the two pre-setup branches (`needsImport`, `!isSetupComplete`) had no `id`, so only the *configured* layout carried anchors. Added `id="identity"` to both — without it the hash lands nowhere, which is precisely the state a user in this toast is in.

## Verification

Reproduced on the preview (identity not yet imported on that device, so the Dawarich tile is in the `requiresSetup` state):

- Toast renders as `Set up your encryption keys before configuring this integration.` + `Set up`.
- Clicking the action lands on `/settings/account#identity` with the Federation Identity section resolved and scrolled to ("Your identity exists on this server, but you need to import…").

Standard sonner action layout — 356×73 toast, 53×24 button — matching the existing action-toast precedent. `vue-tsc` clean; `integrations.store.test.ts` green.

No screenshot attached: the toast only appears in the pre-setup state, which is easiest to see live on the preview link below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Changed
 
+* The "set up your encryption keys" warning on an integration now links
+  straight to the setup screen
+
 ### Fixed
 
 * Dragging the bottom sheet upward and releasing now lands on the next
--- a/web/src/components/identity/IdentitySettings.vue
+++ b/web/src/components/identity/IdentitySettings.vue
@@ -168,6 +168,7 @@ function handleSetupComplete() {
     <!-- Needs Import (has server keys but no local) -->
     <SettingsSection
       v-else-if="needsImport"
+      id="identity"
       :title="t('friends.identity.title')"
       :description="t('friends.identity.description')"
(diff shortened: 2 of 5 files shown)


alexwohlbruck:

Issue #136, opened 2025-04-19
Layer groups


alexwohlbruck:
Layers should be able to be grouped into a parent layer. When this is done, the children are removed from the layer picker and toggling the parent group will toggle all the children simultaneously. This layer group should feel and behave just like other layers, but will simply control multiple layers at once.

Examples to use this with:

Mapillary: Overview/Sequences/Images

Contour lines, labels, hillshade

Loom Transit:Tram/Light Rail/Commuter/Long Distance

PR #494, opened 2026-09-09
Multimodal trip types

alexwohlbruck:
Multi-stop trips become real journeys, transit gains the strategies it was missing, and the per-stop time controls are rebuilt for touch.

## Trip types

Transit directions now offer park-and-ride and bike-to-station **without a vehicle saved in the app** — these are properties of the network, not of your garage. Saving one still improves the result: the trip starts by walking you to where you actually parked.

The combined mode now shows one trip per *kind* rather than several of whichever kind ranks best, so riding to the station and taking a shared bike at the far end are both visible instead of being buried under walk-and-ride departures. Transit-only search still lists alternative departures, since there a shortlist is the point.

A bike ridden to a station is now parked at a real rack and walked in, rather than being silently abandoned at the platform. Bike carry-on is offered where the service allows it (see the barrelman companion below).

## Multi-stop is one coherent journey per strategy

It used to chain one walk-access query per waypoint pair, `numItineraries: 1`, and return a single concatenated trip — one itinerary, no alternatives, none of the strategies above.

Now the first leg's strategies each become a whole journey, extended stop by stop:

```
 26 min  reaches stop 1 at 13:23  driving>walking · driving>walking
 52 min  reaches stop 1 at 13:29  biking>walking  · biking>walking
 65 min  reaches stop 1 at 13:41  walk+transit    · walk+transit
 66 min  reaches stop 1 at 13:38  bike>transit    · walk+transit
155 min  reaches stop 1 at 14:29  walking         · walking
```

Continuity is physical. A leg ending on your own car or bike hands it to the next leg; a leg that parks it ends on foot and hands nothing on — which is why the bike+transit journey walks its second leg, the bike being at a rack by the first station. Mid-journey legs are planned from what you're carrying, not from the vehicles saved at home.

`planChainedIntermodalTrips` is deleted.

## Per-stop times, rebuilt for touch

The clock on each stop was `opacity-0` until hover, so the whole feature was unreachable on a phone. It's now always visible at 40×40 with a real label, showing `From 1:45 PM` when set and a quiet `1:29 PM` when it's only what the trip implies.

The sheet behind it is all taps: a segmented Leave-after/Arrive-by, day chips with a full calendar behind **Other**, a custom scrolling dial for the time, one-tap `+15 / +30 / +1 hr`, and stay-here durations. Nothing re-plans the trip until Done — it used to emit on every keystroke, and a plan takes seconds.

**The dial is custom.** `<input type="time">` is a different control on every platform and none of them match the app. Three snapping columns behave the same everywhere; the band is positioned on the row grid rather than the control's centre, and a `ResizeObserver` re-seats the selection once the column first has a size (scrollTop on a zero-height box goes nowhere).

**Times can't overlap.** `lib/directions/leg-schedule.ts` walks the trip forward for the floor — you can't leave before you arrive, plus a one-minute gap — and backward for the ceiling. Only `arrive by` caps an earlier stop; asking to leave a later stop after some time says nothing about this one, since you can wait. Out-of-range rows dim, dialling past a bound clamps back, day chips disable, and moving a stop later carries any pinned stop after it along.

You can now also ask to stay somewhere without fixing a time to arrive.

## Two bugs worth calling out

**Short trips took 38–60 s.** Instrumented rather than guessed: a 772 m walk was making 16 routing calls, ten of them pricing park-and-ride — driving somewhere you'd walk in nine minutes, then hunting for a space. Each call can stall 6–16 s while GraphHopper faults its memory-mapped graph back in (7–23 ms warm). Park-and-ride now has the same distance floor transit already had. **38–60 s → 1–8 s**, and a plain walking option surfaces where it was crowded out.

**Bike parking could set off before you left.** Riding via a rack takes longer than MOTIS's direct bike leg, and the difference came out of the start with no floor — so a 13:00 departure produced a trip starting at 12:57, and a second leg beginning before the first ended. That boarding is now dropped and the next itinerary gets a turn.

## Also fixed

* Stops appeared after every *segment* rather than at leg boundaries, so a leg that walks to a station and rides on drew its "via" marker mid-leg.
* The constraint picker's mode selector rendered blank — reka's item registry is empty until the list has been opened once.
* Scroll areas inside a bottom sheet could only be dragged one way: the sheet's touchmove guard cancelled downward swipes on their behalf. It now defers to a nested scroller that still has room.
* `views/directions/waypoint-time-constraints.test.ts` reimplemented the logic it tested, which `CLAUDE.md` forbids. The logic now lives in `lib/` and the tests import it. Net −583 lines.

## Companion

Bike carry-on needs [alexwohlbruck/barrelman#136](https://github.com/alexwohlbruck/barrelman/pull/136) deployed **and a re-import** — MOTIS bakes the timetable at import time. Until then Parchment correctly shows no carry-on trips: the server requires the provider to confirm it enforced carriage, and now stops re-asking a provider that can't, rather than spending a MOTIS round trip per plan on it.

## Verification

* **167 server tests, 1744 web tests, 0 failures.** `tsc`: server 266, web 0 — both unchanged from baseline.
* Merged `dev` (trip list view, transit network dimming, POI stop markers). Two files overlapped: the changelog, where both sides filled in the same headings, and `TripDetail.vue`, where dev extracted the place-stop derivation into `lib/directions/trip-stops.ts` — took theirs and rebased the leg-boundary fix onto it.
* Exercised end to end against the preview API, asserting no leg starts before the previous ends and none before the requested departure.
* A review pass over the full diff turned up nine issues, all fixed in `5c41d7c4` — an egress rack detour not updating `latestEndTime`, continuations planning a drive from a car parked two stops back, a failed POI search cached permanently as "no rack here", a false "misses the next stop" warning, a draft time off the dial's five-minute grid, an `arrive by` ceiling that ignored the stay, and stale swagger/docs.

**Not verified on a device.** The mobile shell drops the directions panel below 768 px (loading `/directions` there redirects and strips the waypoints), and the automated browser has no touch emulation — so the dial's feel under a thumb, and the sheet-gesture fix specifically, are untested. Worth a pass on a phone.


alexwohlbruck:

alexwohlbruck:
Preview refreshed with all of the above: https://vega5.tailc47d6.ts.net:8401

PR #495, opened 2026-09-09
Dedupe walk routing

alexwohlbruck:
One trip plan enriches every walk leg of every itinerary across ~5 MOTIS queries, and those itineraries overlap heavily — the access walk to a given station is identical in every itinerary that boards there. Each copy was routed separately.

`cachedWalkRoute` memoises the pedestrian re-route on endpoints plus preferences, mirroring the existing `cachedNearestEntrance` pattern (same FIFO trim, same promise-caching so concurrent callers share one in-flight request). Preferences are part of the key because they change both the route and the language of its instructions. Rejected routes are evicted so a transient failure isn't cached.

## What this is and isn't

**It is** a reduction in redundant load on Barrelman, which is production. Measured on a Brooklyn → Manhattan transit plan: 9 routed walk legs collapse to 5 distinct routes, a 44% cut. The real saving is larger, since enrichment runs over every itinerary, not just the ones that survive filtering.

**It is not** a latency win. Interleaved runs against `dev`: dev 1.69–1.94s, this branch 1.70–2.22s — no meaningful difference. Those calls were already concurrent and bounded by the 48-slot semaphore in `barrelman-integration.ts`, and wall-clock is dominated by the MOTIS queries. Ship this for the load reduction or not at all; it will not make trips feel faster.

## Correctness

The cached route object is now shared across segments, so the leg is read-only at the assignment site (noted in a comment there). I checked for post-enrichment mutation of `geometry` / `instructions` / `edgeSegments`: there is none, and `collapseWalkableTransitLegs` runs before enrichment and shallow-copies.

Output is unchanged. An early comparison appeared to show a dropped bike-share leg, but that was live GBFS availability drifting between two time-separated runs — interleaving the two servers gives byte-identical trip shapes.

## Tests

Two added. `routes each distinct walk once across itineraries` fails on `dev` (4 calls, expected 2) and passes here. The second is a regression guard that a failed route isn't cached; it passes both ways by construction.

Full trip suite 144 pass / 0 fail. `tsc` error count unchanged from `dev` at 320.

### DIFF
--- a/server/src/services/trip.service.test.ts
+++ b/server/src/services/trip.service.test.ts
@@ -2211,6 +2211,61 @@ describe('TripService — scoring', () => {
 
 // ── Validation ──────────────────────────────────────────────────────────────
 
+describe('TripService — walk route deduplication', () => {
+  /** Two itineraries on different lines that share the same access/egress walks. */
+  function makeSharedWalkItineraries() {
+    const a = makeTransitItinerary()
+    const b = makeTransitItinerary()
+    b.legs[1] = { ...b.legs[1], routeId: 'route-7y', routeShortName: '7Y' }
+    return [a, b]
+  }
+
+  test('routes each distinct walk once across itineraries', async () => {
+    mockGetIntermodalRoute.mockImplementation(async () => ({
+      itineraries: makeSharedWalkItineraries(),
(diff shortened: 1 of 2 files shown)


alexwohlbruck:

PR #496, opened 2026-09-09
Map markers for trip stops

alexwohlbruck:
A trip that parks the bike at a rack showed that stop on the timeline but not on the map — the rack was an invisible corner in the line. Now every stop the plan makes on your behalf wears the same marker the origin and destination do.

## What changed

- `lib/directions/trip-stops.ts` — one pure reading of "which segment boundaries are places the plan stopped at". The timeline and the map both use it; the timeline used to walk the segments itself.
- `TripStopsLayer` — a marker layer for those stops, following the same trip as the instruction points (z-index 2: under the waypoints you asked for, over the instruction dots).
- `WaypointMarker` lost `totalWaypoints` (never read) and `index` (only ever meant "not the origin", which `type` already says), and gained `label` (a stop the plan named, like "Bike parking", has no name on its `Place`) and `draggable` — the route-stop markers the map strategies draw were never draggable but wore `cursor-move` anyway.

Transit boarding and alighting stops are not covered: they arrive as `TransitStop` (id, name, location), not a `Place`, so there is no POI glyph to draw. Every timeline entry that carries a `Place` now has a marker.

## Testing

`trip-stops.test.ts` covers the boundary rules — last segment excluded, place-less and location-less boundaries skipped, label fallback. `vue-tsc` clean; 255 tests in the touched areas pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### Added
 
+* Stops a trip makes for you — the bike rack it parks at, the lot it leaves the
+  car in — now show on the map as their own labelled POI marker, matching the
+  timeline
+
 ### Changed
 
 ### Fixed
--- a/web/src/components/map/markers/WaypointMarker.vue
+++ b/web/src/components/map/markers/WaypointMarker.vue
@@ -23,12 +23,16 @@ import type { Place } from '@/types/place.types'
  * at 13px, 1.1em below the icon's centre. A stop is a POI, so it should not be
  * possible to tell which of the two labels the style drew.
  */
-const props = defineProps<{
-  index: number
-  totalWaypoints: number
-  type?: 'origin' | 'destination' | 'waypoint'
-  place?: Partial<Place> | null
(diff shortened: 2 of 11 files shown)


alexwohlbruck:

alexwohlbruck:
**Verified on the preview** (slot 5): the stop draws its own POI marker with a "Bike parking" label, next to the destination's house marker — the same plate/glyph/label treatment the origin and destination get.

Caveat on how it was verified: this slot's API could not plan a live trip (barrelman returned 30s timeouts, then `Connection between locations not found`, for every mode — barrelman's own `/health` is green, so it looks environmental to this slot rather than related to the change). The persisted snapshot from the original trip URL also 404s here, since each preview gets its own database clone. So the trip was injected into the directions store in the browser instead, with a bike-rack `Place` on the first segment's end — which exercises `tripPlaceStops` → `TripStopsLayer` → `WaypointMarker` end to end, just not the planner that produces the segment.

PR #497, opened 2026-09-09
Scroll the trip timeline

alexwohlbruck:
The trip suggestions timeline was clipped: a trip longer than the fitted range (an 11-hour Park & Ride among 40-minute transit trips) ran off the right edge with no way to see the rest of it. `overflow-x-clip` was deliberate — making the list a scroll container would have captured the time axis's vertical stickiness.

## What changed

* The trip rows now live in their own horizontal scroller. The time axis stays outside it — keeping its `sticky top` docking under the pinned controls — and mirrors the scroller's offset with a `translateX`, so ticks and bars stay locked together.
* Everything pans together: bars, and the duration/arrival/distance column. The only thing that pins is each row's summary line (`Transit & Walking · 529 g CO₂` / `4 or 5 · toward Woodlawn`) — it travels with the bar until it reaches the panel's left gutter, then holds there, so the rows stay identifiable however far you scroll.
* Removed the row's `gap-3.5`: the axis grid had no gap, so ticks sat 14px left of the bars they labelled. The gutter now comes from the sidebar's padding, and the axis lines up with the timeline.

Verified in the preview — summary position by scroll offset: `0 → 95px`, `40 → 55px`, `95 → 12px`, `900 → 12px`, while the duration column tracks the content the whole way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@
 ### Fixed
 
 * The trip suggestions timeline scrolls sideways, so a trip that runs past the
-  visible range can be followed to its end. Each suggestion's duration, arrival
-  and summary stay pinned in place while the bars pan.
+  visible range can be followed to its end. Each suggestion's type and route
+  description stay pinned at the edge so you can tell the rows apart.
 
 ## [0.11.5] - 2026-09-09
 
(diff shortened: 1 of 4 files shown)


alexwohlbruck:

PR #498, opened 2026-09-09
All trip types in multi

alexwohlbruck:
**Stacked on #494** — base is `multimodal-access-strategies`, not `dev`. Merge that first; without it, park-and-ride in multi would still require a registered vehicle.

Multi returned one representative per top-level mode but suppressed every *composite* transit strategy, so it answered "which mode?" and never "which combination?" — the question the epic opens with.

## Generation

Two of the three `isMulti` suppressions come off. The distinction is whether a query adds a **type** or only **variants of a type**:

- **Vehicle access** (park-and-ride, bike-to-station) — new types, now run in multi.
- **Rental egress** (transit → shared bike) — a new type, now runs in multi at `numItineraries: 2` rather than 4.
- **Few-transfers sweep** — stays suppressed in multi. It only yields more variants of walk+transit, which is exactly what multi doesn't need.

## Selection

`filterQualityTrips` took the top `MAX_TRANSIT_OPTIONS` (8) transit trips by signature in rank order. In a dense network walk+transit fills all 8, so a park-and-ride was dropped *even when it was the best trip of its type* — the `bestPerMode` exemption only covered the quality floor, never the transit gate.

Now a reservation pass claims one slot per distinct strategy first; remaining slots fill by signature as before. The quality floor is untouched, so a genuinely unviable option is still dropped — "one of each" means one of each that works, not one of each invented.

`getTripMode` also became positional. It keyed on the first non-walk segment, so a bike ridden *to* the station and a dock bike ridden *from* it both returned `biking+transit` and collided on a single reserved slot. The key is now `biking+transit:access` vs `biking-shared+transit:egress`.

## Measured

Brooklyn → Manhattan, multi, interleaved against `dev`:

| | dev | this branch |
|---|---|---|
| trips | 6 | 11 |
| bike-to-station | absent | present |
| transit + shared-bike egress | absent | present |
| latency | 1.56–1.71s | 2.57–2.70s |

**The latency cost is real: roughly +1s, about 60%.** That is the price of two extra MOTIS round-trips plus the rental query. For scale, dedicated transit mode on the same trip is 3.1s, so multi stays the cheaper of the two. If that trade isn't acceptable, the knob is `numItineraries` for multi (currently 3) — say the word and I'll tune it rather than guess at your threshold.

## Two things this does not do

- **No park-and-ride appeared** on this test route. The `CAR_PARKING` query runs and returns nothing for a Brooklyn → Manhattan trip, which is plausibly correct — driving to a station is a poor option there. It means park-and-ride is exercised but not yet *observed* end to end; a suburban origin would test it properly.
- **Variants within a strategy are still uncapped.** Multi now shows three `biking > transit` options. The per-strategy reservation guarantees one; the fill pass adds more by signature, and transit trips bypass `MAX_PER_MODE` (pre-existing). More than asked for, but noisier than a clean overview. Easy to cap if you want multi to be strictly one-per-type.

## Tests

`a lone strategy survives a full slate of better-ranked walk+transit` builds 8 faster walk+transit routings ahead of one park-and-ride; it fails on the parent branch and passes here.

Two `getTripMode` assertions updated for the positional keys. Suite: 147 pass / 0 fail. `tsc` error count unchanged from the parent at 324.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 
 ### Fixed
 
+* Trips that ride a bike to transit now park it at a real bike rack and walk
+  you in, instead of leaving the bike unaccounted for once you board.
+
 ## [0.11.5] - 2026-09-09
 
 ### Added
--- a/server/src/services/trip.service.test.ts
+++ b/server/src/services/trip.service.test.ts
@@ -1626,6 +1626,115 @@ describe('TripService — scoring', () => {
     })
   })
 
+  // ── Bike parking before boarding ────────────────────────────────────────────
+
+  describe('bike parking before boarding transit', () => {
+    const stationRack = {
+      geometry: { value: { center: { lat: 35.2103, lng: -80.8496 } } },
+      name: { value: 'Stop A bike rack' },
+      tags: {},
+    }
+    const bikeVehicle = {
+      id: 'bike-1',
+      type: 'bike' as const,
+      location: { lat: 35.209, lng: -80.861 },
+    }
+
+    /** MOTIS rides the bike to Stop A and says nothing about parking it. */
+    function mockBikeAccessItinerary() {
+      mockGetIntermodalRoute.mockImplementation(async (req: any) => {
+        if (req.preTransitModes?.includes('BIKE')) {
+          const itinerary = makeTransitItinerary()
+          itinerary.legs[0] = {
+            ...itinerary.legs[0],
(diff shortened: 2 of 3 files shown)


alexwohlbruck:

alexwohlbruck:
Folded into #494, which targets `dev` directly. No stacked branches — the commit is unchanged, just carried on the one branch.

PR #499, opened 2026-09-09
Trip list view

alexwohlbruck:
Adds a second way to read trip suggestions: a classic Google/Apple Maps style list, alongside the existing timeline. A toggle next to the sort control switches between them, and the choice is remembered.

Each list row shows the trip's duration, its departure–arrival window, a strip of leg chips (walk minutes › route badge › walk minutes), and the same type / distance / cost / CO₂ meta line the timeline shows.

**Shared instead of duplicated**

The two views were going to restate a lot of the timeline's logic, so it moved out first (first commit, no behaviour change):

* `lib/directions/trip-type.ts` — the trip-combination taxonomy, previously an untyped computed inside `TripItem.vue`
* `lib/directions/trip-chips.ts` — the leg strip
* `TripMeta.vue` — the type / cost / CO₂ line and the headsign line, now rendered by both views
* `TripRows.vue` — ranking, dividers and the staggered entrance
* `composables/directions/useTripNavigation.ts` — opening a suggestion, including the snapshot POST
* `formatTimeRange` in `trip-display.ts`, which keeps both am/pm markers when a trip crosses noon

`TripItem.vue` also loses its duplicate `formatCo2`, which already lived in `trip-display.ts`.

Tests cover the extracted rules: trip type (including that an absorbed wait doesn't count as walking), chip building, and the time range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 * Stops a trip makes for you — the bike rack it parks at, the lot it leaves the
   car in — now show on the map as their own labelled POI marker, matching the
   timeline
+* Trip suggestions can be shown as a plain list instead of the timeline, with
+  each option's legs as a row of route badges. The toggle sits next to the
+  sort control and is remembered
 
 ### Changed
 
--- a/web/src/components/directions/TripItem.vue
+++ b/web/src/components/directions/TripItem.vue
@@ -4,8 +4,9 @@ import dayjs from 'dayjs'
 import { useI18n } from 'vue-i18n'
 import { useMapService } from '@/services/map/map.service'
 import { useUnits } from '@/composables/useUnits'
-import { getTravelModeCssClass, getTravelModeColor } from '@/lib/directions/travel-mode-colors'
-import { getSegmentIcon, getModeIcon } from '@/lib/directions/travel-mode-icons'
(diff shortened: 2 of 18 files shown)


alexwohlbruck:

alexwohlbruck:
**Verification**

`vue-tsc` clean; 105 web tests pass, including the new `trip-type`, `trip-chips` and `formatTimeRange` suites.

Visual check was done in the preview with a stubbed trips payload pushed into the directions store, because barrelman routing was timing out for the whole preview fleet at the time (every freshly-started preview server hits `timeout of 30000ms exceeded` on `getRoute`, so real planning returns "No routes found"). Unrelated to this branch — it changes no server code.

The three stubbed suggestions render as:

```
41 min                              4:00 – 4:41 PM
🚶 6 min › [B45] › 🚶 3 min › [Q] › 🚶 4 min
🚇 Transit · 3.3 km · $2.90 · 169 g CO₂
B45 · toward Downtown Brooklyn
────────────────────────────────────────────────
26 min                              4:00 – 4:26 PM
🚶 2 min › 🚲 23 min › 🚶 1 min
🚲 Bike Share · 2.1 km · $11.73 · 12 g CO₂
```

Route badges take the line's own colour, the toggle button flips its icon and title, and the choice survives a reload. Worth a second look on a real plan once routing recovers.

PR #500, opened 2026-09-09
Park the bike before boarding

alexwohlbruck:
Stacked on #498. Fixes the bike+transit trips being wrong: they ride a bike to the station and then walk away from it, with nothing saying where the bike went.

## The defect

MOTIS's `BIKE` access mode rides to the boarding stop and models no parking event — `BIKE` is a plain street mode, not a park-and-ride mode like `CAR_PARKING`. We passed that through untouched, so the timeline read *cycle 19m → board the F → walk 9m*, a trip nobody can actually take.

## The fix

The bike ride is re-aimed at the nearest `amenity/bicycle_parking` within 250m of the boarding stop, followed by a short walk in:

```
biking (origin → rack) → lock → walking (rack → stop) → transit → …
```

Both new legs are **back-timed off MOTIS's original hand-off**, so the boarding time never moves — the rider sets off earlier instead of missing the train. A 60s lock-up gap sits between the ride and the walk. When a bike is registered, `parkedVehicles` records the rack, so a return trip knows where it is.

Rack lookups are cached per stop (itineraries share boarding stops), and reuse `searchNearbyParking` + the `segmentsConnect` barrier guard that parking-aware driving already uses.

## Two judgement calls worth your input

**Gated on `useKnownParkingLocations`.** That preference is already the contract for "route me to real parking", and it bounds the added search to users who opted in. With it off, the trip stays as MOTIS planned it — still unparked. Flipping this to always-on is a one-line change; say so if you'd rather the trip never lie regardless of preference.

**No rack → the trip is dropped.** There is nowhere to leave the bike, so it isn't a trip you could take. A *failed* lookup is treated differently and leaves the trip alone, so a routing outage doesn't silently delete options.

## Your preferred variant — bike on board, park at the far end

You said "preferably after". That needs bike carriage on the transit leg, and I checked what the data supports rather than guessing:

* MOTIS does have `requireBikeTransport`, and the composition works end to end. Verified against Seastreak (`BIKE > FERRY > BIKE`, 48 min).
* **Only 2 of the 44 loaded GTFS feeds declare `bikes_allowed=1`** — Seastreak and Concord Coach. None of the MTA feeds do, so `requireBikeTransport=true` returns **zero** itineraries for your Brooklyn → Manhattan trip while the same query without the flag returns five.

So carry-on is blocked by feed data, not by routing. Bikes genuinely are allowed on the NYC subway; the feed just doesn't say so. Closing that needs a `bikes_allowed` policy injection in barrelman (same shape as the existing Fares v2 synthesis) plus a passthrough for the parameter — a production data change, so I've left it for you to call. Tracked against PAR-252.

## Verification

* 151 tests pass. The 3 new behavioural tests fail on the parent branch and pass here; the 4th pins the preference-off path.
* `tsc`: 324 errors, identical to the parent baseline — none in the new code.
* **Not yet verified end to end.** Street routing is down on both paths right now: hosted GraphHopper is returning "Minutely API limit heavily violated", and barrelman's own GraphHopper on vega2 is not running, so `/route`, `/directions` and `/isochrone` are all reporting unavailable. Both new legs need street routing, so the preview will show the untouched trip until one recovers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 
 ### Fixed
 
+* Trips that ride a bike to transit now park it at a real bike rack and walk
+  you in, instead of leaving the bike unaccounted for once you board.
+
 ## [0.11.5] - 2026-09-09
 
 ### Added
--- a/server/src/services/trip.service.test.ts
+++ b/server/src/services/trip.service.test.ts
@@ -1626,6 +1626,115 @@ describe('TripService — scoring', () => {
     })
   })
 
+  // ── Bike parking before boarding ────────────────────────────────────────────
+
+  describe('bike parking before boarding transit', () => {
+    const stationRack = {
+      geometry: { value: { center: { lat: 35.2103, lng: -80.8496 } } },
+      name: { value: 'Stop A bike rack' },
+      tags: {},
+    }
+    const bikeVehicle = {
+      id: 'bike-1',
+      type: 'bike' as const,
+      location: { lat: 35.209, lng: -80.861 },
+    }
+
+    /** MOTIS rides the bike to Stop A and says nothing about parking it. */
+    function mockBikeAccessItinerary() {
+      mockGetIntermodalRoute.mockImplementation(async (req: any) => {
+        if (req.preTransitModes?.includes('BIKE')) {
+          const itinerary = makeTransitItinerary()
+          itinerary.legs[0] = {
+            ...itinerary.legs[0],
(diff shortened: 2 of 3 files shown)


alexwohlbruck:

alexwohlbruck:
Not actually merged into anything — GitHub auto-closed this as merged when its commit moved onto `multimodal-access-strategies`. The work now ships via #494, which targets `dev`. Nothing has landed on `dev` yet.

PR #501, opened 2026-09-10
Fade the transit network on trip views

alexwohlbruck:
Applies the route detail page's transit treatment to trips: the surrounding network steps back so the trip's own line reads as the subject, and live vehicles are scoped to the lines the trip actually rides.

## What changed

**Fading.** A trip on the map — hovered in the list or opened — dims the transit network behind it. A trip that rides *no* transit hides it altogether: there the ribbons aren't context for a walking or driving line, just clutter beside it.

Deliberately **not** a portolan isolation like the route panel's. Portolan bundles routes at parallel slot offsets, so lifting the A out of the tiles would draw it *beside* the trip line rather than under it — the same line twice, in two styles. The trip line is already the highlight; the dim is the other half of it.

**Vehicles.** Filtered to the legs' routes (interchangeable lines like the 4/5 included), with the specific run you're catching held at full strength and the rest faded back — reusing the same `selected`/`dimmed` props the route panel's vehicle selection already drives, so the two pages can't drift apart.

The planner names a run with its own encoded token that the realtime feed has never heard of. The departure board names it with the feed's own trip id, which is exactly what a vehicle carries, so `TripDetail` pushes what its board resolved and that outranks the planner's — which also means the highlight follows the train after a rebooking. When nothing can be matched, **nothing dims**: every train at full strength beats the wrong one lit.

## Structure

A shared `transit-focus` store is the single answer to "is the map showing a particular piece of transit, and which vehicles belong to it", asked by the vehicle layer and the network dim, answered by whichever page owns the map. `transit-vehicles-layer.ts` went from three separate `routeDetailStore.isActive` branches to reading that one façade.

Also extracted along the way: the network dim (was inline in `route-isolation.service.ts`) and the route-scoped vehicle fetch (was duplicated in `route-detail.store.ts`).

## Two bugs found and fixed in review

**The dim only ran once.** It walked the style and painted every `portolan-*` layer that existed at that instant, then latched. Portolan mounts ribbons progressively as tiles hydrate, and on MapLibre builds *two* layers per ribbon (solid + ghost, no `line-occlusion-opacity`) — so everything mounted afterward stayed at full strength. Broken on MapLibre; worked on Mapbox only by luck of timing. Portolan now owns the dim the same way it owns route isolation: folded into the opacity expression, so ribbons are *born* dimmed and re-derive it on every rebuild.

**The dim died on the layer toggle.** `teardownPortolanTransit()` reset the flag, so toggling the Transit group off and on rebuilt the network at full strength. The dim is UI intent, not layer state, so it now outlives teardown.

## Verification

Measured on the MapLibre engine through the real app, at three zooms each after a full layer rebuild:

| state | portolan layers | dimmed | hidden |
|---|---|---|---|
| transit trip, z12.4 / z13 / z15.5 | 44 / 40 / 36 | all | 0 |
| Transit group toggled off → on | 88 lines + 5 symbols | all | 0 |
| walking-only trip | 85 | 0 | 85 |
| trips cleared | 85 | 0 | 0 |

`vue-tsc` clean, 2311 tests pass (13 new, covering the feed-prefix handling that decides whether you light up Metro-North's 2 or the LIRR's).

**Not verified:** the `hidden → dimmed` transition directly (hovering from a driving suggestion straight onto a transit one). Same two setter calls `apply()` already makes, but it needs a mixed-mode list and `mode=multi` times out against barrelman on the dev box. Mapbox hasn't been re-measured since the ownership change either — it routes the same `isolationOpacity()` result through `ribbonColorWithAlpha` instead of `line-opacity`, so it should follow, but that's reasoning rather than a measurement.


alexwohlbruck:
Live preview (tailnet): https://vega5.tailc47d6.ts.net:8406 — currently set to the **MapLibre** engine, which is where the one-shot dim was failing.

PR #502, opened 2026-09-10
Hide transit network for non-transit trips; keep dim across layer toggle

alexwohlbruck:


alexwohlbruck:
Duplicate — opened accidentally by `preview pr` from the preview worktree's branch name. The real PR is #501.

PR #503, opened 2026-09-11
Map labels and bike network

alexwohlbruck:
Two passes over the map's cartography, in one branch: the label system retuned
after Apple Maps, and the cycling network repainted onto the roads instead of
over them.

---

# 1 — Apple-style labels

`convert-basemap-style.mjs` grew a `TYPOGRAPHY` pass: a table keyed by layer id,
merged over the converted spec, plus five tokens (`label_halo`,
`label_ink_strong` / `_muted` / `_road` / `_road_minor`). It throws if an id
stops existing **or** if a text layer appears outside the table, so a
regenerated spec can't silently leave a layer in MapTiler's treatment.

Four things carry the hierarchy where size used to do it alone:

| | before | after |
|---|---|---|
| weight | Regular streets, Medium POIs, Regular stations | Medium → SemiBold → Bold |
| ink | near-black everywhere | cool slate for names of things, warm grey for streets |
| halo | hard 1px, no blur | 1.2–1.6px at 0.9 alpha, lightly blurred |
| leading | MapLibre's 1.2 | 1.05 on anything that wraps |

Plus two Apple moves: **streets go to small tracked capitals at z17** and step
back so the POIs carry the map; and **a stop is lettered like a place** — Apple
tints the mark and leaves the name alone, where ours put a station in transit
blue, level with the café next door.

Every `label_ink_*` clears 4.5:1 against what it sits on, both flavors, with a
test that computes it.

### Bugs found on the way

* **Station names were shipping in Regular, not SemiBold.** `basemapLabelFont()`
  took the *first* upright face in the style; on our own basemap that's
  `Oneway`. Now picks the heaviest, extracted to `labelFontFor` and covered.
* **Cycling, timezone and contour labels were not drawing at all on MapLibre** —
  the compat layer produced a two-name Geist stack, which MapLibre requests as
  one `%2C`-encoded path segment. No such directory, labels gone.
* **Route-isolation station labels** were `#333333` on `#ffffff` regardless of
  theme, naming a font the glyph endpoint has never had.

---

# 2 — A green bike surface

A bike map usually draws its network *over* the roads: a dotted line riding the
centreline. At street zooms that is a second object on top of a first — it hides
the road's casing, never quite follows the curve underneath, and a dense grid
reads as hatching. So **the road is painted green instead.**

### How the width match is guaranteed

`addCyclingSurface` generates a twin of each road surface and casing. A twin is
the road layer spread, with the road's filter **narrowed** by a cycling clause
(so a designated way in a tunnel stays untinted, as its road does), the road's
`line-width` and `line-gap-width` **copied by reference**, and one thing of its
own: `line-color`.

A green road is therefore exactly as wide as the road it replaces at every zoom,
because it *is* that road's width — no second ramp to drift, and a later retune
of the roads reaches the twins for free. Tested with `toEqual` per twin.

Qualifying: `subclass=cycleway` or `bicycle=designated`. Deliberately **not**
`bicycle=yes` — permission, not provision, and it's on most of the residential
grid. Our tiles already carry these, so the green surface draws even when
Barrelman is down.

### Slots, and the two things they fixed

MapLibre `addLayer` with no `beforeId` lands above the entire style, labels
included. `layer-slots.ts` honours Mapbox Standard's own three bands:

* `bottom` — above the polygons, **below the roads**
* `middle` — above the roads, below the labels
* `top` — above the labels

On-street lanes take `middle`, which is what puts them under the POI names and
under the transit ribbons that cross above them. The anchor is checked against
the **real shipped style**, not a toy list, including the case where transit is
on — that is where the first attempt went wrong.

The signed-route corridor takes `bottom`, which fixes a second defect: it was a
wide translucent band over the roads, and MapLibre blends each feature
separately, so every corner and junction where two segments overlapped compounded
into a dark blot. No paint property prevents that. Below the roads it can be
**opaque** — nothing to compound — and wider than the street it follows, so what
shows is a green verge either side of a street that stays crisply drawn on top.
A server test now refuses any `middle` layer that isn't opaque.

### The rest of the layer set

`cycling.ts` went 1177 → ~290 lines. Fourteen near-identical configs became one
shape plus a palette; the six layers the basemap now draws better are gone. What
remains is on-street infrastructure: protected track deepest, painted lane a step
back, sharrow dashed because that is honestly what it offers, `bicycle=yes` faint
and off. The white casings — most of what made the old rendering read as hatching
— are gone. The group moved below transit and picked up `basemapGroup`, a
template-only field (no migration) letting an overlay group switch on basemap
layers.

---

## Tests

Web green. New coverage: the weight ladder, halo floor, leading, the capitals
switch arriving as one step, the two ink families staying apart, 4.5:1 per
flavor, `labelFontFor` incl. the `Oneway` regression, the cycling width-match
per twin, what the tint selects and rejects against real feature shapes, slot
anchors against the shipped style, and the band/opacity wiring server-side.

Server has one pre-existing `dev` failure (`personal-blob.controller`),
untouched here.


alexwohlbruck:

alexwohlbruck:
### Verified on the preview

Driven through both flavors and both zoom regimes on the MapLibre engine (the
Parchment basemap — note that if your device's map engine is set to **Mapbox**,
none of this is visible, since Standard brings its own cartography).

| scene | camera |
|---|---|
| light, mixed-case streets | `-73.9782, 40.6835` @ z15.4 |
| light, capitals switch | `-73.9905, 40.6860` @ z17.9 |
| dark | `-73.9782, 40.6835` @ z15.4 |

What the three show:

* **z15.4 light** — "Atlantic Avenue–Barclays Center" bold slate over three
  tight rows; "Fulton Street" / "Lafayette Avenue" / "Bergen Street" the same;
  street names medium grey along the line, clearly a rung below; "Nicholas
  Naquan Heyward Jr. Park" green and wrapping into one block; "GOWANUS" in
  muted tracked capitals.
* **z17.9 light** — BERGEN STREET / WYCKOFF STREET / WARREN STREET / SMITH
  STREET in small tracked grey capitals, with the POI names ("Penguin Coding
  School", "Elite Gourmet & Deli", "Mosaic House") SemiBold and tinted above
  them. This is the switch, and it reads the way the reference does.
* **z15.4 dark** — near-white station names, light-grey streets, green park,
  dark halos. The hierarchy survives the inversion.

No screenshots attached: the repo is public and the review surface here is the
live preview, which is a better look at label placement than a still anyway.

PR #506, opened 2026-09-11
Green bike surface

alexwohlbruck:
Stacked on #503 — it builds on that branch's basemap tokens and its
`TYPOGRAPHY` pass. Retarget to `dev` once #503 merges.

## The idea

A bike map usually draws its network *over* the roads: a dotted line riding the
centreline. At street zooms that line is a second object on top of a first — it
hides the road's own casing, it never quite follows the curve underneath, and a
dense grid reads as hatching rather than as streets you can ride. The second
screenshot in the request is that failure mode at city scale.

So **the road is painted green instead**.

## How the width match is guaranteed

`addCyclingSurface` in `convert-basemap-style.mjs` generates a twin of each road
surface and casing layer — `Minor road`, `Major road`, `Highway`, `Path`,
`Path bridge` and their outlines. A twin is the road layer spread, with:

* the same source, source-layer and geometry
* the road's filter **narrowed** by a cycling clause (not replaced — so a
  designated way in a tunnel stays untinted, exactly as its road does)
* the road's `line-width` and `line-gap-width` **copied by reference**
* one thing of its own: `line-color`

So a green road is exactly as wide as the road it replaces at every zoom,
because it *is* that road's width. There is no second ramp that can drift out of
step with the first, and a later retune of the roads reaches the twins for free.
A test asserts the width expressions are `toEqual` their base's.

Each twin splices directly above the layer it derives from, which means casings
stay under surfaces, a green minor road still draws below a major road, and the
whole set sits below every label and everything any overlay draws.

## What counts as cycling infrastructure

`subclass = cycleway` or `bicycle = designated`. Deliberately **not**
`bicycle = yes` — that is permission, not provision, and it is on most of the
residential grid; tinting it paints whole neighbourhoods green while saying
nothing about where it is good to ride. Tested against the real OpenMapTiles
property shapes, both the matches and the misses.

Our basemap tiles carry `bicycle`, `subclass`, `foot` and `surface` on
`transportation`, which is what makes this possible without a second source —
and means the green surface still draws when Barrelman is unreachable.

## The rest of the layer set

`cycling.ts` went from 1177 lines to ~280. Fourteen near-identical configs (each
with its own hand-written `measure-light` pair and width ramp) became one
`wayLayer` shape plus a palette, and the two halves the basemap now draws —
`cycleways` and `bicycle-paths`, six layers and two subgroups — are gone.

What's left is on-street infrastructure, which is *not* the whole road and so is
not tinted: a protected track takes the deepest green, a painted lane a step
back, a sharrow is dashed because that is honestly what it offers, and
`bicycle=yes` is faint and off by default. Signed routes became a soft wide
corridor *behind* the network rather than another green line competing with it.
The white casings are gone — they were most of what made the old rendering read
as hatching.

## Ordering

Both complaints had one cause: MapLibre `addLayer` with no `beforeId` puts a
layer on top of the entire style, labels included.

`layer-slots.ts` honours the slot vocabulary Mapbox Standard already has. A
layer that says `slot: 'bottom'` is ground, and gets an anchor that keeps it
there — before the first basemap label **or** the first `portolan-` layer,
whichever comes first, so a bike lane also stays under the train line that
crosses above it. Only `bottom` is emulated; `middle`/`top` already mean "above
the basemap" on MapLibre, and giving them anchors would move layers that are
correct today.

The cycling group also moved below transit (`order: 3`), and picked up
`basemapGroup: 'cycling'` — a template-only field (no migration) that lets an
overlay group switch on basemap layers, which is the only place cartography this
exact can be drawn.

## Tests

Web green: **2543** across 146 files. New coverage: the width-match guarantee
per twin, splice position, ship-hidden, filter narrowing, what the filter
selects and rejects against real feature shapes, the toggle finding every twin,
and six cases on slot anchoring including the overlay-label trap.

Server has the one pre-existing `dev` failure (`personal-blob.controller`),
untouched here.

## Not yet seen on a map

The browser automation channel dropped partway through, so **I have not looked
at this rendered** — the reasoning and the tests are solid but the colour values
are unverified by eye. The preview is up; `cycling_surface` / `cycling_casing`
in `tokens.*.json` are the two dials if the green is off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,16 @@
   name that wraps reads as one block instead of separate lines
 * Street names shrink to small spaced capitals as you zoom in, so the shops and
   stations you're looking for carry the map
+* Cycleways and paths where bikes are designated now show as a green road
+  surface rather than a dotted line laid over the street, exactly as wide as
+  the road itself at every zoom
+* Bike lanes, tracks and sharrows are redrawn as thin markings along the street
+  they're painted on, with each form a different weight so a protected track
+  reads apart from a shared lane
+* Signed bike routes now show as a soft corridor behind the network instead of
+  another green line competing with it
+* The cycling layer now draws under place names and under transit lines, so a
+  train crossing a bike lane passes over it
 
 ### Fixed
 
@@ -30,6 +40,7 @@
   font it has never had, and silently drew nothing
 * Station names on a line's own page now follow the map's theme instead of
   always being dark text in a white halo
+* Bike route labels no longer sit on top of the places you're looking at
 * The trip suggestions timeline scrolls sideways, so a trip that runs past the
   visible range can be followed to its end. Each suggestion's type and route
   description stay pinned at the edge so you can tell the rows apart.
--- a/server/src/constants/default-layers/cycling.ts
+++ b/server/src/constants/default-layers/cycling.ts
@@ -4,1118 +4,254 @@ import { LayerType } from '../../schema/layers.schema'
 /**
  * Cycling default layer templates.
  *
- * Convention: layer `order` values are spaced in intervals of 10 within each
- * subgroup (110, 120, 130, …) so that user-driven reorders can slot new
- * templates or clones between existing ones without immediately needing a
- * full re-index. When adjusting values, keep the spacing consistent so the
- * client-side sort stays predictable.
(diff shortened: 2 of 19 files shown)


alexwohlbruck:

PR #507, opened 2026-09-11
Quick edit

alexwohlbruck:
Prototype for PAR-152: add and edit OpenStreetMap POIs directly in the app, using the user's connected OSM account.

## What works
- **Add a place**: map context menu → Edit → *Add a place*. Draggable pin, preset search over the iD tagging schema (1 709 presets, localized names/terms), dynamic form generated from the preset's field definitions, always-available raw key/value tag editor.
- **Duplicate guard**: before adding, nearby elements with the same primary tags are fetched from Overpass and offered as "edit this instead".
- **Edit a place**: pencil action on the place panel (anything with an `osm` external id) or via a duplicate suggestion. Prefills from the **live** OSM element, submits a tag diff; nodes can be moved by dragging the marker. Ways/relations are tags-only.
- **Opening hours editor**: structured weekly grid (multiple intervals per day, copy-to-all, 24/7) for the common grammar subset; values it can't round-trip are edited as raw text and never rewritten.
- **Changesets**: create → upload → close per edit, `created_by=Parchment <version>`, user-supplied comment with sensible default. Conflicts (element changed upstream) return 409 with the live element and the form reloads.
- **Pending edits**: submissions are recorded (`osm_edits` table) and surfaced as a "pending map update" notice when re-editing, since the imported map data lags OSM.

## Notes
- Edits go to whatever server the OpenStreetMap **system** integration selects — the dev DB points at the sandbox (`master.apis.dev.openstreetmap.org`), so this preview is safe to submit from. Verified end-to-end against the sandbox: created node 4365663236 (changesets 677532–677535), including a stale-version conflict.
- Fixed en route: public Overpass rejects axios's default User-Agent with HTTP 406 — all Overpass/OSM API calls now send an app User-Agent.
- Out of scope for the prototype: way/polygon geometry editing (desktop-only iD-style editing discussed for later), deletion (will become lifecycle-tag "closed" action), AI autofill (PAR-296), address autofill from reverse geocoding, `es-ES` strings.

Linear: PAR-152


alexwohlbruck:

alexwohlbruck:
Screenshots from the preview (slot 1, OSM sandbox):

**Preset search** — type anything, matched against localized names/terms of the iD schema:

preset search

**Add form with duplicate guard** — pin dropped 7 m from an existing Starbucks:

add form

**Editing an existing element** — prefilled from the live OSM node (the one this branch created on the sandbox), pending-edit notice, structured hours parsed from `Mo-Fr 07:00-17:00`:

edit existing

alexwohlbruck:
### Brand lookup (Name Suggestion Index)

Chains now resolve to their canonical OSM tagging, the same way iD does — `brand`, `brand:wikidata`, cuisine, takeaway and whatever else the chain always carries — plus the brand logo.

**Typing a name suggests chains** (ranked by how widely the chain operates, so "taco" reaches Taco Bell before a two-country chain with a shorter name):

brand suggestions

**Picking one fills everything in** and the logo replaces the preset icon:

brand applied

**Editing a sloppily-tagged chain offers the fix**, reproducing iD's "looks like a common feature with nonstandard tags" card with a tag diff:

brand match

#### Notes
- NSI's own `Matcher` costs **~320 MB RSS** to index, so this builds a smaller index from the same inputs (simplified names, match groups, per-tag exclusion lists). Matching still handles case/spacing (`taco bell`, `TACOBELL`), cross-tag matches (a chain catalogued under `amenity/fast_food` but tagged `amenity/restaurant`), and refuses generic names (`pizzeria`, `Imbiss`).
- Still **~130 MB resident**, built lazily on first brand use. Worth precomputing a trimmed index at build time before this ships widely — filed as follow-up, not done here.
- Indexes the `brands` tree only; `operators` is larger and covers utility/transit networks no one adds through quick edit.
- 10 unit tests cover ranking, normalization, match-group and exclusion behaviour, and the tag diff.

alexwohlbruck:
### Form redesign, address autocomplete, POI icons, alias search

**Fast Food went from ~3 screens to one.** Sections are collapsed by default and open themselves when they already hold data:

form

**Yes/no tags are tri-state chips.** Seven labelled button pairs became two rows, and they can now express "not answered" — which is the state most of them are actually in:

features

**Address autocomplete** through our own geocoder, plus "Use pin's address" (reverse):

address

**"bike rack" → Bicycle Parking**, with the app's own POI icon:

alias

**Duplicates are drawn on the map**, de-emphasised, and the map frames them:

duplicates

#### Notes
- Aliases were **server-side** all along (`server/src/services/category.service.ts`). Preset search now goes through that same index, so quick edit and app search agree, and presets arrive with `iconName`/`iconPack`/`iconCategory` already resolved.
- The Yes/No bug was real: `toggleVariants` defaults to a solid `bg-primary` base and selection only adds a pale `data-[state=on]:bg-accent` — inverted. Quick edit no longer uses ToggleGroup; **other callers (e.g. the isochrone mode switch) still have the inverted look** and are worth a separate fix.
- Dependent fields hide until relevant, via the schema's own `parent/child` id convention plus a short explicit map (drive-through hours need a drive-through). 12 unit tests.
- `SearchResultMarker` was refactored onto a shared `PoiMarker`, which quick edit reuses for the pin and the muted duplicates.
- Fixed en route: `stashPath(to.path)` dropped query strings, so any deep link with params (notes, quick edit) lost them while auth resolved.
- **Not done:** hiding the basemap POI layer during quick edit — there's no "off" for it today (`poiStyle` is `badge|glyph`); the muted markers read clearly enough without it.

alexwohlbruck:
### Tag suggestions + form chrome

**Yes — the schema package can back it.** `@openstreetmap/id-tagging-schema` ships `taginfo.min.json`: 976 documented keys and 3,223 key=value pairs, plus every field's `options`. No taginfo.io calls, no new dependency. The raw editor now suggests keys, and values scoped to the key you're on, with arrow-key navigation. Free text still wins — suggestions are a shortcut, not a whitelist.

tag suggestions

**Closed sections now say what they hold** instead of showing a bare number — the single biggest readability win, since a count tells you something's inside but not whether you need to open it:

previews

Also in this pass:
- Section rows are ~40% shorter, each with an icon, and content is indented under it so the hierarchy reads without dividers between every row.
- Raw tag rows are `key = value` with an aligned `=`, monospace, on one line.
- Tag edits commit on blur or selection rather than per keystroke — renaming a key was previously rewriting the tag map letter by letter.
- Lighter footer: the change note lost its redundant label, and the sandbox warning is one muted line instead of a filled block.
- 34 unit tests across grouping, previews, address mapping and hours.

alexwohlbruck:
### Wiki links per field

Every field label now carries a help icon linking to the tag's OpenStreetMap wiki page, and so does each row in the raw tag editor.

wiki links

It resolves the destination the same way iD does:
- A tag with a discrete value gets its own page — `cuisine=mexican` → `Tag:cuisine=mexican`.
- Free-text fields (name, phone, website) link to the **key** page; there's no wiki page per phone number.
- Fields documenting a tag family use the `reference` the schema declares — the address field's `addr:*` resolves to `Key:addr`.
- Multi-values and wildcards fall back to the key page.

Opens through `useExternalLink`, so it behaves in the Tauri builds. 8 unit tests on the URL resolution.

alexwohlbruck:
### Named markers

The pin now carries the place's name under it, styled the way the basemap labels its own POIs — category-coloured, 11px medium, white halo (dark in dark mode), centred under the plate and click-through so it can't block dragging.

The duplicate markers are labelled too, which is arguably where it matters most: "Raising Cane's / Taco Bell Cantina / Jollibee" on the map answers *"is mine already here?"* far better than three identically-iconed dots.

Added to the shared `PoiMarker`, so search-result markers can opt into labels later with one prop.

**Implementation note:** Vue root props aren't reactive, so a mounted marker can't be told a new name — the pin is re-placed when the name changes, debounced 400ms so it doesn't remount on every keystroke. While doing that I collapsed the element view's two marker-creation sites into one `placeMarker()`.

Verified against the running preview (the camera won't settle reliably in a fresh headless profile, so this is DOM/computed-style rather than a zoomed screenshot):
```
{ text: "taco bell", color: "rgb(255, 153, 51)", fontSize: "11px", fontWeight: "500",
  textShadow: "rgba(255,255,255,0.9) 0 0 3px, …", pointerEvents: "none",
  visible: true, gapBelowPlate: 2, centred: true }
```

**Worth knowing:** these are DOM markers, so unlike the basemap they don't declutter — at low zoom with several duplicates the labels can overlap. At real editing zooms (17-19) with a handful of candidates it's fine, but if you want it bulletproof the duplicates would need to move to a symbol layer with collision detection.

alexwohlbruck:
### Fuzzier brand lookup

**Cause was an inconsistency in my own code.** `matchBrand` (the "this looks like Taco Bell" card) normalized through NSI's `simplify()`, which strips punctuation, case and spacing — so it already caught `mcdonalds`. But `searchBrands` (the type-ahead) was doing plain lowercase substring matching against the display name, and `"mcdonalds".includes` never reaches `"McDonald's"`. Two paths, two notions of a match.

Both now score through `simplify()`:

| typed | finds |
|---|---|
| `mcdonalds`, `McDonalds`, `mcdonald's` | McDonald's |
| `raising canes` | Raising Cane's |
| `ben and jerrys` | Ben & Jerry's (`&` → `and`) |
| `tacobell` | Taco Bell (spacing dropped) |
| `seven eleven` | 7-Eleven |

Two further changes, both matching what iD does:
- **Alias names are searched**, not just display names — NSI's `matchNames` plus the chain's own `name`/`brand`/`operator` tags, scored slightly below the display name so the canonical entry still ranks first.
- **Search spans the match group.** Previously, picking "Restaurant" for a McDonald's surfaced nothing, since NSI catalogues it under fast food. The chosen tag still wins ties. `matchBrand` already did this; search now agrees.

Memory is unchanged (~136 MB for the index, lazily built) — the added per-brand alias arrays are simplified strings that were already being computed for the exact-match map.

3 new regression tests, 13 on the NSI module total.

alexwohlbruck:
### Chains in the type search, compact hours, linked changeset

**Typing a business name in the place-type search now just works.** Picking a chain sets the place type *and* its tags in one step:

chains in picker

Chains are listed **below** place types, so "cafe" still leads with the Cafe type while "mcdonalds" — which matches no type — lands straight on the chain. Only strong matches qualify, so a loose hit on a common word can't bury the type you're after.

Two supporting fixes:
- `searchBrands` can now search **every** chain, not just one primary tag, since no preset is chosen at that point.
- Dedup was picking the chain representative by reach, ignoring score — so "mcdonalds" surfaced *McDonald's PlayPlace* (same QID, wider reach) instead of McDonald's. Score now decides; reach only breaks ties.

**Opening hours**, roughly a third of the height it was:

hours

- One line per day instead of a time row plus a separate "Add hours" row.
- Times use the app's own `TimePicker` — `<input type="time">` is a different, oversized control on every platform, which is what made it so tall.
- Days you haven't set read "Closed" and are one click to open.
- Add-interval and copy-to-all stay hidden until you hover the row.

**Toast** now says just "Saved to OpenStreetMap" with a **View** action opening the changeset on whichever OSM server the edit went to.

Also fixed: brand logos showed an empty box while loading (lazy, no placeholder) — the initial now shows until the image arrives, and stays if it never does.

PR #509, opened 2026-09-12
Reconnect OSM account

alexwohlbruck:
## What

Three things broke connecting an OpenStreetMap account:

1. Reconnecting failed with `Integration openstreetmap-account is already configured with scheme server-key`, and the account couldn't be disconnected either.
2. Quick edit's "connect your account" screen had no button.
3. After a successful connect, the OAuth popup stayed open showing a second copy of the whole app, with the success toast inside it.

## Why

**(1)** `getConfiguredIntegrations` drops rows whose config can't be decrypted, so an account stored under a retired `PARCHMENT_INTEGRATION_ENCRYPTION_KEY` is invisible to the OAuth callback. The callback saw nothing to remove, inserted a fresh row, and collided with the stale one on the `(user, integration, scheme)` unique index. Disconnect failed the same way, with a 404 — leaving no route out of the state from the UI. Live on production: every integration row written before the key changed on 2026-07-17 is unreadable there.

**(2)** The action was passed through `EmptyState`'s `#action` slot, which doesn't exist — actions go in the default slot — so it rendered nothing.

**(3)** openstreetmap.org serves `Cross-Origin-Opener-Policy: same-origin`, so the popup loses `window.opener` the moment it navigates there. The callback page's `postMessage` could never arrive, and it fell through to its "no opener" branch: redirect the popup to `/settings/integrations`. The service worker then served the app shell for that navigation — hence a whole second app inside the popup.

## Changes

* `deleteUserIntegrations(userId, integrationId)` removes a user's rows for an integration straight from the table, readable or not. The OAuth callback and disconnect both use it.
* The callback now hands its result to `/oauth/osm.html`, a static page on the *client* origin. Being same-origin with the app it broadcasts the result over a `BroadcastChannel` and closes itself; the tile listens on that channel instead of `message`.
* `/oauth/` is excluded from the service worker's navigation fallback, so that page is served as itself rather than as the app shell.
* Both quick edit views share one `ConnectPrompt`, whose button deep-links to `/settings/integrations?integration=openstreetmap-account` — that seeds the search so only that tile shows.

## Verified

* `deleteUserIntegrations` removes an undecryptable row a listing can't see — new DB-backed test.
* With the OSM row's ciphertext corrupted, `POST /integrations/osm/disconnect` returns 200 and clears the row (it 404'd before).
* Driving the real flow on the preview: `opener after OSM: false`, `POPUP SELF-CLOSED: true`, `TOAST IN APP WINDOW: true`.
* Screenshots in the comments.

## Notes

* Rows that don't decrypt are still invisible everywhere else in settings, so the stale system-integration rows on production can't be cleaned up from the UI. Not fixed here.
* `/oauth/osm.html` is a plain static file, so its two lines of copy aren't run through i18n.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
   configured" error.
 * Quick edit's "connect your account" screen now has a button through to the
   OpenStreetMap integration in settings.
+* Connecting an OpenStreetMap account no longer leaves a second copy of the app
+  open in the sign-in window — it closes itself, and the confirmation appears in
+  the window you started from.
 
 ## [0.12.0] - 2026-09-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 ### Fixed
 
+* Reconnecting an OpenStreetMap account no longer fails with an "already
+  configured" error.
+* Quick edit's "connect your account" screen now has a button through to the
+  OpenStreetMap integration in settings.
+
 ## [0.12.0] - 2026-09-11
 
 ### Added
--- a/server/src/controllers/osm-oauth.controller.ts
+++ b/server/src/controllers/osm-oauth.controller.ts
@@ -8,7 +8,7 @@ import { db } from '../db'
 import { tokens } from '../schema/tokens.schema'
 import {
   createIntegration,
-  deleteIntegration,
+  deleteUserIntegrations,
   updateIntegration,
   getConfiguredIntegrations,
 } from '../services/integration.service'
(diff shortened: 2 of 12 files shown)


alexwohlbruck:

alexwohlbruck:
### Connect prompt (account not connected)

Connect prompt

### Where the button lands

Settings deep link

alexwohlbruck:
### OAuth popup handoff

Connecting no longer leaves a second copy of the app in the popup: the popup closes itself and the toast lands in the window you started from.

Toast in the original window

Measured on the preview, driving the real tile → popup → handoff flow:

```
popup at OSM:        https://master.apis.dev.openstreetmap.org/login?co
opener after OSM:    false     <- OSM's COOP: same-origin severs it
POPUP SELF-CLOSED:   true
TOAST IN APP WINDOW: true
```

alexwohlbruck:
### Manual callback on previews

The console workaround was gated on `window.location.protocol === 'http:'`, so it never armed on a preview — those are HTTPS, so the popup path ran and dead-ended on the OAuth app's registered `https://localhost:5000/...` redirect.

The server now says whether it can expect the browser back: `/integrations/osm/authorize` returns `manualCallback`, true when the registered redirect URI's origin isn't this server's. The callback also content-negotiates, so the manual path reads JSON instead of scraping HTML.

Driven on the preview with a real state token and a deliberately bogus code:

```
CALLBACK RESPONSE 200 {"status":"error","message":"OAuth2 error: invalid_grant"}
TOAST: OAuth2 error: invalid_grant
```

State validated, exchange attempted, error surfaced in the app — a real code completes it.

Production is unaffected: its redirect URI is `https://api.parchment.app/integrations/osm/callback`, which matches `SERVER_ORIGIN`, so `manualCallback` is false and the popup flow stands.

PR #510, opened 2026-09-12
Unblock the release build

alexwohlbruck:
## What

`vue-tsc` fails, so `bun run build` and `bun run build:ci` can't complete. That took the v0.12.0 release down: Android, iOS and desktop all failed, so **Create GitHub Release**, **Upload to App Store** and **Upload to Google Play** were skipped ([run 34654918384](https://github.com/alexwohlbruck/parchment/actions/runs/34654918384)). The tag and the server shipped; nothing else did.

## Why

`web/src/types/quick-edit.types.ts` imported its types from `@server/services/osm-edit.service`. That service imports `integration.service` → `services/integrations` → every adapter, so a type-only import pulled the server's whole integration graph into the frontend's typecheck and surfaced 37 pre-existing type errors in code the browser never runs (valhalla, overpass, wikipedia, geoapify, graphhopper…).

Those errors aren't new — `tsc` in `server/` reports 266 of them, and has for a long time, because nothing gates it. What was new in 0.12.0 was the web build being made to look at a slice of them.

## Change

The four shared shapes move to `server/src/types/osm-edit.types.ts`, a leaf module with no runtime imports. The service and the controller import them from there; the web app does too.

## Verified

* `bunx vue-tsc --noEmit` — 0 errors (was 37).
* `bun run build:ci` — completes.
* `bun run test` — 2452 passed.
* `tsc` in `server/` — still 266 errors, unchanged. This PR doesn't fix them, it stops the web build depending on them.

## Follow-up

The server's own typecheck being red is worth its own pass — some of those errors look real (`valhalla-integration` reads `avoidHills` off a type that only has `avoidTolls`). Nothing runs `tsc` over `server/` in CI, so they'll keep accumulating.

### DIFF
--- a/server/src/controllers/osm-edit.controller.ts
+++ b/server/src/controllers/osm-edit.controller.ts
@@ -17,8 +17,8 @@ import {
   submitEdit,
   getPendingEdits,
   OsmEditError,
-  type OsmElementType,
 } from '../services/osm-edit.service'
+import type { OsmElementType } from '../types/osm-edit.types'
 import { integrationManager } from '../services/integrations'
 import { IntegrationId } from '../types/integration.enums'
 import type { OverpassIntegration } from '../services/integrations/overpass-integration'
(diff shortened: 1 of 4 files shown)


alexwohlbruck:

PR #514, opened 2026-09-12
Trip line stops

alexwohlbruck:
An itinerary's transit legs were drawn as bare coloured lines: no stations, no
names, no bullets. The network behind them dims, so what a leg calls at was
harder to read on a trip than on the route panel.

The symbols now narrow to the trip the same way they narrow under a route
isolation, over several lines at once. Each leg keeps its own stations at full
strength while the ribbons stay dimmed, their names and connection bullets come
with them, and the caterpillar bullets ride the span the leg actually rides —
not the rest of the line. Where portolan draws the leg that is portolan's own
lettering; where it does not, the stops fall back to the same
circles-and-names overlay the route panel already uses for a feed with no
pyramid.

### Notes

* `portolan-transit.service` gets `setTripLegs`, the multi-route sibling of
  `setIsolatedRoute` — it narrows the symbols only, never the ribbons, because
  portolan draws bundled routes at parallel slot offsets and lifting the leg's
  ribbon would land it beside the trip line rather than under it.
* Stations match a leg by its own stop list; bullets, which ride between stops,
  match the chain of them (`distanceToLine`, new in `lib/geo/geo-line`).
* The route panel's circles-and-names overlay moves to
  `transit-stop-overlay.ts` so both callers draw a stop the same way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+* Trip maps now draw the stops, station names and line bullets along each transit leg, instead of a bare coloured line.
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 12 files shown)


alexwohlbruck:

PR #517, opened 2026-09-14
Stable 3D building heights

alexwohlbruck:
With 3D terrain enabled, buildings near vector-tile borders rendered as two interleaved copies standing at different ground heights, and changed elevation as the map zoomed.

## Cause

Two independent things, both rooted in how a building's ground elevation is sampled:

1. **Tile-border duplication.** A vector tile duplicates a building into every tile whose buffer it touches, clipping it at the buffer edge. Flat, the copies are coincident and depth testing hides the overdraw. With the terrain mesh up, MapLibre stands each copy on the DEM at the centroid of the geometry *as clipped into that tile* — different tiles, different centroids, different elevations. The copies separate vertically and interleave.
2. **Zoom-dependent elevation.** MapLibre 6 defaults `zoomLevelsToOverscale` to 4, re-parsing tiles into re-clipped deeper canonical tiles past the source maxzoom. Every split level cuts buildings along a brand-new grid, so each new grid hands a building a new centroid and a new DEM elevation — it changed height once per zoom level, at a zoom depending on its distance from the camera.

## Fix

**`zoomLevelsToOverscale: undefined`** on the map (`maplibre.strategy.ts`) restores pre-v6 overscaling, so one bucket and one clipping serve every close zoom and a building keeps a single elevation. This is the revert documented in MapLibre's own [v5→v6 migration guide](https://github.com/maplibre/maplibre-gl-js/blob/main/docs/guides/v5-to-v6-migration-guide.md), not a hack — upstream notes it also slightly affects `queryRenderedFeatures` and polygon-center label placement, i.e. we are back on the v5 behaviour we shipped against before.

**`web/scripts/patch-maplibre-terrain.mjs`** then reconciles the borders of that one remaining grid, with two per-polygon rules inside `FillExtrusionBucket.addFeature`:

- *Ownership* — a tile only draws a polygon whose bounding-box centre it owns, half-open over `[0, EXTENT)` so a centre exactly on a border belongs to exactly one tile. Drops buffer duplicates. Bbox centre rather than vertex mean, so a copy is only dropped when the neighbour's copy fully covers it.
- *Border anchoring* — a polygon crossing a tile edge line is a clipped piece whose counterpart crosses the same world line through the same wall geometry. Both compute the midpoint of their crossings of that line and sample the DEM there instead of at their own clipped centroid: same anchor, same elevation, overlap renders coincident. A polygon crossing both an x- and a y-line is corner-spanning, where span midpoints measured on each piece's own clipped geometry would disagree across the diagonal — those anchor at the crossed corner point itself, the one world position all four pieces share.

Ported from subwaybuilder.

## Why a postinstall patch

The fix lives inside a bucket method in a minified single-line bundle, so a `.patch` file would be a diff of the entire file. The script is a targeted string patch instead, run from `postinstall`, and is idempotent. It asserts each anchor appears **exactly once** and fails the install loudly otherwise, so a `maplibre-gl` upgrade that moves the code stops the build and gets the patch re-derived rather than silently dropping it.

## Verification

Against the installed `maplibre-gl@6.4.1-transit.1`, all four anchor strings (prologue site and centroid-override site, in both `maplibre-gl-shared.mjs` and `maplibre-gl-shared-dev.mjs`) match exactly once. `FillExtrusionBucket` lives only in those two shared chunks — the worker and main bundles import it rather than carrying their own copy — so both the dev and production paths are covered.

Needs a look on the preview with terrain on: buildings astride tile borders should sit at one height, and hold that height across a zoom sweep.


alexwohlbruck:

PR #518, opened 2026-09-14
Dark mode shields

alexwohlbruck:
Highway shields went white on the night map again, and reappeared at every zoom.

## What happened

Three commits tuned the shields in September — dark sprite variants (2ad0acc), zoom-gated density and the junction drop (c003955), and the spacing ramp (4bc9c01, 298f392). All of them edited `spec.json` and `tokens.dark.json` by hand.

But those two files are **generated**: `web/scripts/convert-basemap-style.mjs` writes them from the vendored MapTiler style. The converter never learned about any of the shield edits, so the next regeneration — f308af9, "Retune map label typography after Apple Maps" — overwrote all of them:

* `@shield_variant` dropped out of `icon-image`, so the style went back to naming the light art (`us-state-2` instead of `us-state-2-dark`) — white plaques over the dark basemap
* dark `shield_ink` reverted to near-black, so the numerals on those plaques were black on white
* the junction exclusion and the by-zoom class gate were replaced by the converter's plain class filter, and the spacing ramp went back to the wider default

The sprite still ships every `-dark` marker; nothing was drawing them.

## The fix

Both treatments now live in the converter, so regenerating reproduces them:

* `shieldImageExpression()` appends `@shield_variant`, and the token is emitted per flavor (`''` / `'-dark'`) alongside the dark `shield_ink`
* `shieldRoadClassFilter()` holds the zoom ramp; the junction exclusion and the spacing ramp move into `routeShieldLayer()`

`spec.json` and both token files are regenerated output — `bun run build:style`.

## Tests

Two in `map-style.test.ts`, both verified to fail on the regressed spec:

* the dark flavor names a suffix that resolves to art the sprite actually carries, for every marker
* a primary road carries no marker at z6 but does at z12; a junction never does


alexwohlbruck:

alexwohlbruck:
The art each flavor now names, straight out of the sprite the style points at — `us-interstate-2`, `us-highway-2`, `us-state-2`, `default-3`, `motorway-exit-2` in the day cut, and the `-dark` twins below:

Shield art, day and night

Before this branch, the night map drew the top row.

**No live map screenshot** — the preview can't draw a basemap at all right now: the Barrelman API key in the dev database is revoked on production. `GET https://api.barrelman.dev/health/auth` answers `401 {"error":"Invalid or revoked API key"}`, so every freshly-started server here fails `barrelman` init and the tile proxy answers 501. The previews that still draw tiles cached their config at boot, before the key was revoked — restarting any of them will take the basemap out there too.

PR #522, opened 2026-09-17
Night woodland

alexwohlbruck:
Night woodland was invisible. MapTiler's dark landcover is a set of blues at the ground's own lightness — woodland `hsl(203, 47%, 22%)` under ground `hsl(216, 37%, 24%)` — so a county that is half forest read as bare land. This gives the planted family the green family the daylight map already uses, carried by hue rather than by lightness: each fill lands within a few percent of the ground's luminance, so woods read without the map turning into a slab.

| before | after |
|---|---|
| before | after |

At street zooms the creek corridors, cemeteries and parks come back without crowding the streets: [z13](https://raw.githubusercontent.com/alexwohlbruck/parchment/assets/dark-woodland/after-z13.png).

### Where the trees come from

They are OpenStreetMap, not a filled-in landcover raster. The `landcover` layer's `class=wood` arrives as one multipolygon per tile — 1,475 rings in the z9 tile over Charlotte, which is why it looks painted on — and 35 of 40 sampled low-zoom wood points are still wood at z14, where the data is raw OSM. Overpass returns `natural=wood` ways at those coordinates.

There *is* a filled-in landcover set, and we drop it: MapTiler's tiles carry `globallandcover` (forest, tree, crop, scrub, grass) at **z0–z8 only**, and the converter drops the four layers that read it because planetiler does not build that source layer for barrelman. Below z9 our map has no landcover at all. Adding it is a barrelman data change, not a style one.

### Review

The preview cannot draw a basemap — the dev database's barrelman key is rejected by production (`401 Unauthorized` on every tile proxy request), so previews render blank. The screenshots above come from the real style rendered against schema-identical OpenMapTiles tiles.


alexwohlbruck:

PR #524, opened 2026-09-17
Terrain camera snap

alexwohlbruck:
## The jump

With 3D terrain on, letting go of a pan that crossed an elevation change makes the whole ground pop up or down.

That is MapLibre's terrain gesture model, not our code:

- `handler_manager.ts:619-626` captures the elevation of the point grabbed at gesture start
- `handler_manager.ts:671-681` sets `elevationFreeze = true`, so the per-frame ground clamp in `map.ts:4299` is skipped for the whole gesture — the pan is solved on one frozen elevation plane
- `handler_manager.ts:747-756` unfreezes on release and calls `recalculateZoomAndCenter` (`transform_helper.ts:613`), which holds camera *altitude* constant and re-solves centre and zoom against the real elevation, in one frame

Our 1.2× exaggeration scales the correction, and a flick pays it twice: the inertial ease re-freezes and `_finalizeElevation` (`camera.ts:824`) recalculates again when the glide stops. Late-arriving DEM tiles can add a third shift.

## This change

Both recalculation call sites are gated on `getCenterClampedToGround()`, so terrain now turns the clamp off with it. The camera holds altitude above sea level, there is nothing to correct on release, and nothing snaps.

The trade-off to judge on the preview: zoom stops meaning height above ground, so low over tall terrain the camera can end up inside the hill. If that reads badly, the alternative is patching our maplibre-gl fork to ease the gesture-end correction instead of snapping it — keeps the clamp, more to maintain.

## Review

Turn 3D terrain on, pitch the camera, and pan across a slope — hills, a coastline, a river valley. Then check the same spots at high zoom over tall terrain for the camera-inside-the-hill case.


alexwohlbruck:

PR #525, opened 2026-09-17
Cycle marks over bridges

alexwohlbruck:
The cycling network lost its mark wherever a way crossed a bridge. The basemap draws a path bridge in a band above the whole road network, and the cycling layers sit below it — so the green line ran up to the deck, vanished across it, and picked up on the far side, reading as a gap in the network rather than as a bridge.

Tunnels had the opposite fault. Nothing filtered on one, so an underpass was drawn exactly like a surface way, on top of the street it passes beneath — and the on-street tint went further, painting a green ribbon where the basemap draws no road at all, since the road layers the tint sits on exclude tunnels.

## What changed

Two slots, `bridge` and `tunnel`, for the thing `bottom`/`middle`/`top` cannot say: which brunnel a way has.

* **On a deck.** Cycleways and bike paths that are a bridge are drawn just above the basemap's decks, so the deck is drawn first and the mark runs along it — the crossing still reads as a bridge. Ways at grade stay where they were, under the decks, so a footbridge crossing *over* a cycleway still covers it.
* **In a tunnel.** Ways in a tunnel are drawn in the basemap's own tunnel band, so the street above draws over the crossing and the way visibly dips under.
* Both carry the stroke alone, no casing: the basemap has already cased the way on its deck and in its bore, and a second casing rubs out the crossing the mark is drawn on. A deck and a bore are both hard-surfaced, so the paved/unpaved dash has nothing to say about them either.
* The on-street tint and the lane markings no longer paint a way that is in a tunnel, matching the rule the basemap's own cycling twins already follow.
* On Mapbox Standard, which has neither slot, they resolve to the nearest band it does have.

Where a band ends is read off the layer filters rather than off layer ids, so a restyle that renames a bridge layer cannot strand an anchor. The bridge anchor names a basemap layer specifically: an anchor is a position and overlays are inserted at it, so one that answered "whatever sits here now" would name the layer added last and stack every further mark underneath it in reverse.

## Before / after

Central Park, rendered against MapTiler's tiles with every footpath painted as a cycleway. Barrelman tiles 501 on a fresh preview, so the shots come from the offline harness rather than the preview.

**Bridges** — the mark breaks at every deck; now it runs along the deck, which still reads as a deck.

| before | after |
|---|---|
| bridge before | bridge after |

**Tunnels** — the underpass drew straight across Center Drive; now it passes under it.

| before | after |
|---|---|
| tunnel before | tunnel after |

## Tests

`layer-slots.test.ts` covers the bands against a hand-written stack, `map-style.test.ts` against the style we actually ship, and `cycling.test.ts` (server) checks each way layer is in the band its brunnel puts it in, that the surface layers leave bridges and tunnels alone, and that a deck and a bore carry no casing.


alexwohlbruck:

